PR #4, opened 2020-12-11
WIP: VTable auto-naming

daemitus:
Biggest blocker: needs to implement vtbl inheritiance so the base classes get named first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -351,3 +351,6 @@ MigrationBackup/
 
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
+
+# JetBrains
+.idea/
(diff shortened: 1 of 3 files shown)


aers:
Do you have ideas for how you want to do the inheritance thing?

I _think_ msvc linker will always have base classes before anything that inherits them so its possible just not re-renaming stuff we've already renamed will handle a lot of it, but we do need a way to reference other vtbls so we don't have to repeat function names multiple times.

daemitus:
@aers So I pulled most of my inheritance out of the old RTTI exe using classinformer and explicitly defined it.  It ends up showing alot of stuff like this:
Warning: vtbl_Component::GUI::AtkComponentInputBase[1] had a name of "Component::GUI::AtkUnitManager_vf1_nullsub"
Warning: vtbl_Component::GUI::AtkComponentTreeList[1] had a name of "Component::GUI::AtkUnitManager_vf1_nullsub"
Warning: vtbl_Client::Graphics::Kernel::Texture[2] had a name of "Client::Graphics::Render::ShadowCamera_vf2"
Warning: vtbl_Component::GUI::AtkComponentNumericInput[1] had a name of "Component::GUI::AtkUnitManager_vf1_nullsub"

New inheritance? Dunno, but its interesting. How the inheritance is actually shown in the big list is a little... busy? I'm tempted to yank it out into its own block, but as of right now it works pretty good. I'd like to get a few people in to test it. If someone knows the ghidra bindings I can fill those in too.

daemitus:
I also suppose this needs testing against ida 7.2 and lower.

aers:
There is definitely new inheritance since 2.3, although you might just be running into classes having multiple vtables.

Like

class A

class B : A

class C

class D : B, C

will have two vtbls, one for D : B : A and one for D : C

daemitus:
well for any inheritances we can actually verify, just need to add it to the list of its parents, for that given class. Otherwise itll just complain, but not overwrite.

PR #6, opened 2020-12-21
WIP: The cycle of improvement continues

daemitus:
- Class renaming has been improved? If a parent class has a non-default name and the child class doesn't, the child override will inherit the name. This is highly dependent on parent classes that have vtbls, having a defined vtbl address.
- Inheritance has been simplified by hand, before classes were listed as class D: A,B,C; class A: B,C. Now it can be written as class  D: A. The resolver will work its way back up the tree. Recursion is magical.
- Made some educated guesses for base classes that were missing vtbl ea entries. Seems to work, needs verification.
- Structs are now being created that match the vtbl, however we are not converting the vtbl to the struct (yet). Running the script twice nukes the vtbl offset entries due to the structs being cleared out if they existed prior. Add it to the todo list, for now people can assign the struct manually. 
- Placeholder classes do not require a 0x0 entry for the vtbl anymore. The ctor is a little ugly, but takes *args and parses out a few sane variations. 
- vtbl 0x141696198 Client::Graphics::Scene::Human, entry vf61@0x141696380 half the time is defined as qword_140444F00 instead of sub_140444F00. Initial auto-analysis doesn't pick it up as a procedure. Run it twice it does get picked up, for now I'm just letting the script print out a warning and continue. At some point this should be handled.

```
(248, 3) # TODO: Clobber 0x141697D60! "Common::Lua::LuaState", "Common::Lua::LuaThread"
(524, 23) # TODO: How to handle this? Only observed instance so far is to unresolved code. 0x141696198 Client::Graphics::Scene::Human, 0x141696380 qword_140444F00, sub_140444F00, vf61
(758, 78) factory.register(0x141659488, "Component::GUI::AtkEventListener", [], {})  # TODO: Verify this
(775, 83) factory.register(0x14166BD10, "Client::Graphics::ReferencedClassBase", [], {})  # TODO: Verify this
(845, 84) factory.register(0x141685A50, "Client::Graphics::Render::BaseRenderer", [], {})  # TODO: Verify this
(967, 76) factory.register(0x14169A3C8, "Component::GUI::AtkEventTarget", [], {})  # TODO: Verify this
(1139, 77) factory.register(0x1416A9CD0, "Client::UI::UI3DModule::MapInfo", [], {})  # TODO: Verify this
```


daemitus:
ok, CullingJobOpt definitely didnt inherit from atk anything. So I'm not sure how that ended up there. Just checked the old RTTI data.

daemitus:
The overloaded function name fetching needs more work.

PR #8, opened 2020-12-27
NodeType and ComponentType,

daemitus:
When the ULD objects were moved, ComponentType got deleted for some reason. I still use it in my UIDebug fork, so I know it works at least for my purposes. 

Also, lets make the Structs contain the enum rather than the raw value. if we're going to compare it anyways, just save a step.

Although ComponentType might be a short instead of an int, dunno.

### DIFF
--- a/Component/GUI/AtkComponentBase.cs
+++ b/Component/GUI/AtkComponentBase.cs
@@ -1,10 +1,10 @@
-﻿using System.Runtime.InteropServices;
+using System.Runtime.InteropServices;
 
 using FFXIVClientStructs.Component.GUI.ULD;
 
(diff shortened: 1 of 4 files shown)


aers:
ULDObjectInfo is a base class for ULDWidgetInfo/ULDComponentInfo, the componenttype is in ULDComponentInfo.

daemitus:
Well that would do it then. I had zero clue going into this when I was initially trying to fix the uidebug stuff.

aers:
It's fine, I meant to finish reworking the UI debug thing after I finished all the component node types, I just never got around to it!

PR #10, opened 2020-12-29
SDK Style projects, addons, fixes, and most importantly, C-style header file exports of structs.

daemitus:
Changed projects to the new SDK style format.

Added a few functions inside event handlers I was poking at.
Added a few addons I was poking at.

C style header exporter for IDA debugging.


daemitus:
All set now.

PR #39, opened 2021-03-12
Some addons

daemitus:


daemitus:
This EventThing as I'm calling it looks to be new since RTTI. If you've got a better namespace/home for it be my guest and I can update before merging.

aers:
Honestly have no idea what it is so I'll just leave it that, shrug.

PR #40, opened 2021-03-16
More addons

daemitus:


daemitus:
Found AtkComponentButtons.Flags at the end of the struct. Appears to change when hovered, held, active/inactive, checked when part of a checkbox.

PR #84, opened 2021-08-02
Add AddonItemSearchResults and AddonRetainerSell

chalkos:


Caraxi:
You probably shouldn't completely destroy the whole Component identifier just to say 1004 is window

PR #134, opened 2021-10-16
Add beast tribe functions to data.yml

rvazarkar:


daemitus:
Can you add these as entries to 
```  Client::Game::UI::PlayerState:
    funcs:
      0x140789110: SetCharacterName
      0x14078CA30: Initialize
      0x1407F2BF0: ctor
```

daemitus:
@aers   Client::Game::UI::PlayerState: is the same as Client::Game::UI::UIModule::PlayerState right? If so, I assume we should fix the namespace.

aers:
Anything with "UIModule::PlayerState" is probably an accident/mistake, it should be Client::Game::UI (even though this is a class with no vfuncs, it definitely doesn't belong to UIModule)

daemitus:
Sorry, not uimodule, uistate. https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/UI/UIState.cs

aers:
Oh. I think its probably in Client::Game::UI and not inside UIModule, yeah.

PR #141, opened 2021-11-09
support multiple inheritance in idarename

aers:
note: also removes generating vtbl structs since I plan to move that elsewhere

This is pretty cursed and currently requires editing ida config files to allow using <>.* etc in names so it doesnt get Very Mad about them. Alternative is to generate mangled names but shrug.

idauser.cfg:

```
#ifdef __PC__
NameChars =
        "$?@*<>,:"           // asm specific character
        "_0123456789"
        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
        "abcdefghijklmnopqrstuvwxyz",
        // This would enable common Chinese characters in identifiers:
        // Block_CJK_Unified_Ideographs,
        CURRENT_CULTURE;
#endif
```

Should be enough.

Otherwise I'm not really happy with the code but its 7 am and I can't be bothered to think better in trees right now. I'll fix it, maybe.

data.yml format has obviously changed a bit. Shouldn't be too hard to decipher, look at Character/BattleChara.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -146,13 +146,12 @@ functions:
   0x14031AC10: SubmitConstantBufferUpdate
 
 classes:
-  # Unknown classes old RTTI data says known classes inherit from
   Client::Game::Control::TargetSystem::ListFeeder:
   Client::Game::InstanceContent::ContentSheetWaiterInterface:
   Client::Game::Object::IGameObjectEventListener:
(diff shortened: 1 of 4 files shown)


aers:
immediately after doing this and getting in bed I thought of the way I really wanted to do it so I'll fix it tomorrow lol

PR #148, opened 2021-12-08
Update  BLMGauge to use flags

daemitus:
@Tischel opinions?


Tischel:
> @Tischel opinions?

Looks good

PR #149, opened 2021-12-08
Fixing offsets for party list addon (trust member count up from 3 to 7)

Tischel:
Opened the PR for visibility but did it as a draft because I wasn't able to fully test this change.


Tischel:
Was able to do some testing on this, seems to be correct.

PR #151, opened 2021-12-10
Add AgentAttribute

Caraxi:
Doing this in a PR because I'm not sure if you'd want it

Added Agent attributes to identify agent classes for easier parsing programmatically


aers:
This is fine, the goal is to simplify using game classes, after all.

PR #156, opened 2022-01-05
erroneous offset in TrustMembers

catgirlfighter:
couldn't get Trust2, and Trust6 would get a chocobo. Noticed that index of Trust6 goes over the size of a struct, so there is an obvious erroneous offset


daemitus (approved):
makes sense

PR #162, opened 2022-01-30
Fix handling of yaml files with empty entries

alandtse:
Fixes a bug if data.yml contains empty nodes under globals, functions,
or classes.


aers:
👀

PR #175, opened 2022-04-09
Add AgentMcguffin to data.yml

KazWolfe:
Adds support for AgentMcguffin to data.yml

This agent controls display of Collection items and the Aether Compass.


daemitus:
lmao. its actually called that.

PR #191, opened 2022-05-02
C exporter fixes

mistahmikey:
Fixes to allow CExporter to run to completion.  Required changes to a few of the existing c# structure definitions, typically to deal with union issues - so these changes may impact other code that I am unfamiliar with.   I also added detail to the InventoryItem struct as regards the InventoryID, which could be removed if it is breaking.  Although CExporter will now finish successfully, I still needed to edit the final header file before ida would successfully compile it.  For a reason I can't uncover, it doesn't like the forward definition of Client::Game::Event::EventHandler, so I had to comment it out.


daemitus:
going to take a look at this now.

Issue #195, opened 2022-05-06
CExporter issue: AddonContextMenu


daemitus:
```
namespace FFXIVClientStructs.FFXIV.Client.UI;
public struct AddonContextMenu
{
    [FieldOffset(0x0)] public AtkUnitBase AtkUnitBase;
    [FieldOffset(0x160)] public unsafe AtkValue* AtkValues; // Does this belong in base?
    [FieldOffset(0x1CA)] public ushort AtkValuesCount; // // Does this belong in base?
}
```

aers:
Yeah those are the arguments passed to the addon open function and should be present in all AtkUnitBases.

PR #197, opened 2022-05-06
Update AtkUnitBase and AddonContextMenu

daemitus:


daemitus:
RE #195

aers:
Double check Dalamud doesn't use this since it was from the context menu

daemitus:
It does
https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Game/Gui/ContextMenus/ContextMenu.cs#L215

aers:
I dunno when goat plans to pull clientstructs again but a note will need to be made of this

PR #203, opened 2022-05-15
Add TargetSystem.AccessObject

marimelon:


aers:
What does this function do? Bit too busy to check it myself right now

pohky:
> What does this function do? Bit too busy to check it myself right now

from the looks of it start interacting/talk with the npc

marimelon:
Start talk NPC, Open Retainer, Open MarketBortd,Open Aetheryte and Others.

It is called when you right-click on an object or when you
When you select a target and press decision on the gamepad, it is called.

aers:
Ah okay thanks.

PR #210, opened 2022-05-18
Update ConfigOption.cs

ItsBexy:
Correction to config names for Expanded Hold settings.
398 and 421 are configs for the Right->Left bar
399 and 422 are configs for the Left->Right bar


Caraxi:
They are wrong because square named them wrong, the names in the enum are the exact names the game uses to save/load

ItsBexy:
That's my bad then; if these name/ID pairs are what's used internally then I would absolutely agree with staying aligned to that.

I DID stumble on other strange namings as well. For example, "HotbarCrossDisp" toggles the "Cross Hotbar Display Type" radio buttons, while "HotbarCrossDispType" toggles "Enable Cross Hotbar" checkbox. I would imagine there's tons more examples throughout.

Maybe the best approach is to stick to SE's naming conventions even when they're bewildering, and just separately document/comment what each option actually does, for cases like the above where the name is misleading or counterintuitive.

PR #247, opened 2022-09-04
Maps out vtbls for 6.2 Addons

juntalis:
Not sure how useful some of these addons are - mostly just did it to help in identifying other bits.


aers:
I need to update data.yml to the latest hotfix before merging this. I'll try to get around to it tonight.

juntalis:
Now that I think about it, I'm not quite sure if I created these mappings before or after the latest hotfix. I'm going to go ahead and close this PR for now until I can verify that they're still correct.

PR #251, opened 2022-09-12
[WIP] Even More Island Sanctuary Work

KazWolfe:
Pushing up as a draft for now; *not verified.* Too tired to continue so here we go for now

- Restore spaces over tabs because every other file uses spaces.
- Add AllowedVisitors bitfield.
- Add tracking of the player's current mode.
- Add tracking of the player's quest progression.
- Add backing bitfields for unlocked key items and recipes.
- Add MJIWorkshops struct for workshop tracking.
- Add MJIGranaries struct for granary tracking.
- Add cabin information.
- Add landmark placement and notes.
- Fix offsets on MJIBuildingPlacement.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/MJIManager.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/MJIManager.cs
@@ -12,7 +12,7 @@ public unsafe partial struct MJIManager {
     /// </summary>
     // Not actually sure about the accuracy of this name. It's a guess based on the fact that the map system and target
     // system appear to change their behavior when this is set to 1, plus verification with how it looks in game.
-    [FieldOffset(0x06)] public bool IsPlayerInSanctuary;
+    [FieldOffset(0x06)] public byte IsPlayerInSanctuary;
 
     /// <summary>
     ///		Represents the currently allowed visitors to the Island Sanctuary.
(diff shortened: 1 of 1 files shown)


KazWolfe:
Some various notes so I have something for tomorrow/whenever:

* Offsets for `MJIBuildingPlacement` and `MJILandmarkPlacement` came from `40 55 56 48 83 EC 68 0F B6 EA` and `40 53 57 48 83 EC 58 80 79 03 00`, respectively. They seem to be shifted a bit earlier than I think we expected, and I'm still not totally sure of everything they contain.
* I'm unsure my use of `vtbl` is right in `MJIWorkshops`/`MJIGranaries`, but that seems to be the convention (?) used in other places. Similarly, I have no idea if my `LayoutKind.Sequential` shenanigans there will actually work; though I would hope they would.
  * I also neglected to add these structs to `data.yml`, though I'm not particularly sure they're important enough? Probably should, honestly.
* The inner struct Pohky mentioned in the initial MJI PR is *definitely* there and is very much real (and annoying). For now I haven't bothered splitting it out, mostly due to a lack of concern plus ease of reading offsets from the disassembler.

aers:
Using a void* for a vtbl placeholder is fine.
You might have to set a Pack on the sequential structs to get them to be right. You'll want to test if they have correct data in-game tbh.

KazWolfe:
Things seemed to test okay, so marking this as ready for review so others aren't blocked if they want more.

@juntalis, gonna poke you here because I had to change `MJIBuildingPlacement` and `MJILandmarkPlacement` a bit, might be good to have some extra eyes on those specifically. I honestly haven't been able to crack 'em fully yet, they behave a bit oddly/frustratingly.

juntalis:
For sure - will check out the changes in an hour or two when I have a sec to hop on.

Also if you feel like including it in your changes, it looks like the byte at offset 0x2C contains the current "Groove" value for the workshops.

PR #257, opened 2022-10-18
Add some extra research for Hotbar things

KazWolfe:
- Some extras for Kami and another sig because why not
- Fix "IsEmpty" to check the same way the game does
  - Old IsEmpty has become IsLoaded, with docs


KazWolfe:
If this doesn't make it in by patch, I'll just close this PR and repurpose it. I have some other things coming for gearsets as well that will require a bunch of `data.yml` updates.

PR #262, opened 2022-11-22
OnSetup, AgentCraftActionSimulator, and an update to RetainerManager

JeffyW:


aers (commented):
Since OnSetup is a virtual function, you can use the VirtualFunction attribute instead of requiring a signature for them.

PR #280, opened 2023-01-12
Fix ExecuteMainCommand member offset

MidoriKami:


MidoriKami:
I don't know what other functions have shifted around, 189 clearly seemed to be an inventory function, but I am not entirely certain what 188 is doing, so I didn't adjust any others.

PR #284, opened 2023-01-12
Fix a few offsets in UI3DModule

Pilzinsel64:
Fix offset of `NamePlateObjectInfoPointerArray` and `NamePlateObjectInfoCounter`
Thanks to Banana Bread for finding them!


aers:
Those offsets changed because the object array was expanded again (to 599) in order to support 8 person cards.

So don't merge this until we can update the rest of the offsets too.

I'll try and do it tomorrow if no one else does.

Pilzinsel64:
Ahhhh, that makes sense! Well, I'm not able to research that at all, so I'll leave that open for now. Just close it if not needed anymore.

Pilzinsel64:
Might this be the same as for `RaptureAtkModule` and `RaptureAtkModule.NamePlateInfo`?
As this also seems to be broken somehow.

aers:
Yes that offset nearly guaranteed to have shifted too. I'm going to bed right now so I'll fix it tomorrow if no one else has by then.

Pilzinsel64:
Alright, thanks for the info! Have a good night and sleep well!

PR #285, opened 2023-01-14
Add AgentGatheringNote

MidoriKami:
Based on AgentRecipeNote
It doesn't seem this function requires the add 0x10_000 that RecipeNote requires.

~~I don't know how to build ClientStructs so I hope this works =x~~
I think I managed to get it to build? It really didn't like the syntax until it was built.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -4921,7 +4921,9 @@ classes:
   Client::UI::Agent::AgentGatheringNote:
     vtbls:
      - ea: 0x141997360
-       base: Client::UI::Agent::AgentInterface
+       base: Client::UI::Agent::AgentGatheringNote
+    funcs:
+      0x1409B47C0: OpenGatherableByItemId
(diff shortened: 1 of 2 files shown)


Caraxi:
Probably shouldn't be called `OpenRecipe`

ida/data.yml
```diff
@@ -4921,7 +4921,9 @@ classes:
   Client::UI::Agent::AgentGatheringNote:
     vtbls:
      - ea: 0x141997360
-       base: Client::UI::Agent::AgentInterface
+       base: Client::UI::Agent::AgentGatheringNote
```
Caraxi, line 4924: This was correct, it should be AgentInterface as base

MidoriKami: Oh I see, because AgentGatheringNote doesn't inherit AgentInterface, makes sense. One moment.

PR #293, opened 2023-01-18
add versioning

Caraxi:
Uses GitVersion to automatically version the assembly using the commit number.

PR'd for feedback in case this is undesirable


aers:
This is a good idea and I don't have  problem with it.

Issue #288, opened 2023-01-16
Future Work


aers:
Generators
=======

- **Static VTables** 
Make vtables available as addresses without needing an instance of the object first. Also provide vtable size via attribute argument. Would be integrated with the virtual function generator and generate a properly-sized vtable struct.
- **Inheritance**
[Inherits<T>] attribute that generates struct members from the parent into the child so all fields and methods show as available without having to do `AtkTextNode->AtkResNode` for example. This generator would need info from all other generators so first combine all generators into a single generator (relatively easy) beforehand.

IDA
===

- **Reformat data.yml**
There are some constructs we don't really support well. Might switch off yaml to something "simpler" since we don't need yaml features.
- **CExporter**
Will likely rewrite this using Roslyn syntax analysis rather than reflection. In addition to exporting types, export method signatures. 
- **Rewrite import script**
Import types and use method signatures to set function argument types appropriately. Also want to pull in some other forms of smart renaming (rename EXD getters via the root, maybe rename unknown agents and addons automatically with their IDs and uld names, more ideas here???)

Other
=====

Anything I'm forgetting. Feel like we've talked about other stuff before.

I feel like some way to indicate which structs have been 'validated' after a patch (anything that didnt change size is probably safe, and anything that changed size and has had its offsets verified) would be nice. There's no doubt numerous "broken" (outdated) offsets and having some way of tracking that would be useful.

wolfcomp:
> * CExporter
> Will likely rewrite this using Roslyn syntax analysis rather than reflection. In addition to exporting types, export method signatures.

This recently received quite the rework and is better at pulling all the data now instead of just struct types. This also included a rewrite of the import method since the file type and layout changed from using c++ headers to a YAML layout

> * Rewrite import script
Import types and use method signatures to set function argument types appropriately. Also want to pull in some other forms of smart renaming (rename EXD getters via the root, maybe rename unknown agents and addons automatically with their IDs and uld names, more ideas here???)

IDA now has the `fullrun` script that runs the `data.yml` import then `exdgetters` then `structimporter` so the only extra addition would possibly be `unknown agents and addons`

wolfcomp:
Closing this as it would be better to keep the one points that are in here that hasn't been touched into a separate issue.

That being: **Reformat data.yml**

PR #300, opened 2023-01-21
Generate static addresses for VTables / VirtualFunctions

JeffyW:
Generate static addresses for VTables, largely using the same technique as StaticAddresses.  If a class has a VTable generated and VirtualFunctions, also generate static addresses for them (that simply offset the VTable).

Very much a work in progress, and I know it's on your list (#288), was hoping this was either close to what you were thinking or was at least food for thought.


aers:
Yeah that's pretty similar to what I had planned.

I started working on my rewrite to better support inheritance generation, but I'm going pretty slow. It will probably add this kind of support as well.

So we can merge this feature now just to have it, if you want.

JeffyW:
Sorry for the delay, needed to get some time to actually test it out (and find a bug).  Also added some concrete addresses and VirtualFuncs that I used personally to confirm it's all working, and updated the README.

There are a few things I'd like to point out in case you have feedback:
- The static VTable is a nuint that points to the VTable.  Wasn't sure if I should define this as nuint*.
- Because it's a nuint, the VirtualFunction addresses both cast and dereference.
```*((nuint*)AddonRetainerTaskAsk.Addresses.VTable.Value + 47);```
- I couldn't quickly come up with a solution I really liked for determining if the parent struct of a `[VirtualFunction]` node was annotated with `[VTableAddress]`.  What I used was:
```syntax.AttributeLists.Any(al => al.Attributes.Any(a => (a.Name as IdentifierNameSyntax).Identifier.Text == "VTableAddress")```

If you're good with all that, then feel free to merge.

JeffyW:
Update, I realized that there's no need to create a VirtualFunctionPointers class, as we can just re-use the VTable class defined for the VirtualFunctions (and this avoids having to do math).  So instead, I define:
```csharp public static AddonRetainerTaskAskVTable StaticVTable => *(AddonRetainerTaskAskVTable*)StaticAddressPointers.VTable;```

This does make it *slightly* more awkward to consume, as the virtual function is now a `delegate*` instead of a nuint, but it can be casted.

I'd love to find a way to use these delegate* as generic pointers, so we could make an overload of
`Hook.FromAddress<T>(T* pointer, T delegate)` that avoids needing to manually define a delegate signature, and enforces the correct one, but I'm not sure it's possible.

aers:
Hey, sorry I haven't been around for a bit to look at this.

Currently working on my combined-generator to support inheritance and I'll likely lift most of this for it, but I'll merge this for now so we can use it already.

> I'd love to find a way to use these delegate* as generic pointers, so we could make an overload of `Hook.FromAddress<T>(T* pointer, T delegate)` that avoids needing to manually define a delegate signature, and enforces the correct one, but I'm not sure it's possible.

Yeah, not possible with how C# is currently constructed. I thought about this before too since it would be a much cleaner way to do it. Maybe some future feature will let it work.

JeffyW:
No prob, thanks for merging!

FWIW, I thought about this a bit and was thinking of trying to make Address generic based on the type of thing the address pointed to.  So, something like: `public partial void OnSetup(uint a2, AtkValue* atkValues);`
Would generate: `Address<Action<nint,uint,nint>>`

And then we could make a Hook.FromAddress<T>(Address<T>, T delegate)

No rush, but let me know if you have any thoughts on such an approach (or just wait till I submit something 🙂)

PR #304, opened 2023-01-22
Even More Island Sanctuary Things

KazWolfe:
This change seeks to move MJI resources into their own folder/namespace. The old MJIManager file was getting pretty large and honestly a bit unwieldy with substructs, so it seemed time to move things into their own folder. Unfortunately, this is a breaking change to plugins that use these APIs. I'm particularly concerned about some of the pointers to _other_ structs in `MJIManager`, as some of those structs are pretty large and/or complex.

If desired, I can drop the rename change and I can instead just add new structs (when they're discovered) to the MJI directory, but might as well do it all in one fell swoop.


KazWolfe:
Re the data.yml removals as I think the commit message got buried:

The removed lines are all duplicates of other lines in the same section. The lua lines are straight duplicates while the EXD lines had multiple names pointing to the same addresses. I elected to keep the more descriptive EXD names, as they also had argument info attached to them.

PR #305, opened 2023-01-22
Add TerritoryInfo

MidoriKami:
`off_14207B5B0`

This struct is created in GameMain_Initialize, it has a size of 0x60 in memory, though most of it seems to always be zero.
I use this struct in `Where Am I Again?` it contains two PlaceName values for area, and sub area.
It seems the Map Agent references these values in a few places.

I'm not sold on the name, if anyone has a better idea for the name let me know.

It's not a manager of anything, so I don't think it's appropriate to be in FFXIV.Client.Game, so I tossed it into FFXIV.Client.Game.UI as it does relate to things like display names and toasts.

image


MidoriKami:
Ah... oops, accidentally pushed the MobHunt struct to the same PR >_<
I'm not going to touch it, I might break something =x

PR #307, opened 2023-01-25
Add WondrousTailsData

MidoriKami:
It didn't seem like this struct was a part of anything I was able to identify.

The red bytes are those that make up the WondrousTailsData struct.
The only pointer that I found points to an address outside of the exe, that results in all ?? in cheatengine.

image


Haselnussbomber:
The address (0x142104BF2) the signature points to is PlayerState+0x63A! 😉

MidoriKami:
Oh awesome, I'll add a reference to it from PlayerState then

MidoriKami:
Holding off until a few questions can be answered.

Hasselnussbomber noted that the functions I tried to put in WeeklyBingo are PlayerState member functions, but I don't think it's a good idea to pollute the namespace that badly.


Ideally I would like to be able to do something like this:
```cs
    [ someannotation ("E8 ?? ?? ?? ?? 3B C3 0F 93 C0")]
    [some stuff here?] uint GetFlagsValue(PlayerState* playerState, uint mode);

    private uint GetFlagsValue(uint mode) => GetFlagsValue(PlayerState.Instance(), mode);
```

MidoriKami:
Closing this PR, as this is apparently not actually a struct, and instead is actual members to playerstate.

PR #310, opened 2023-01-27
Festivals / Events

AsgardXIV:
This is what determines/changes the active festival (Moonfire or whatever).


AsgardXIV:
This just covers the actual layout changes. 
The quest availability and such is calculated in GameMain and I'll submit a PR for that in future once I clean it up.

PR #320, opened 2023-02-04
Move GetBeastTribeAllowance to QuestManager

Haselnussbomber:
The static function `PlayerState.GetBeastTribeAllowance()` is misplaced because the passed private `GetBeastTribeAllowancePointer()` is actually `QuestManager_GetSingleton()`.

I have added it as a member function to QuestManager and marked `PlayerState.GetBeastTribeAllowance()` as obsolete.


MidoriKami:
This is a breaking change, leave the old property and mark it as obsolete with an error.

MidoriKami:
Disregard my now deleted message, I mis-read the PR'd code.

I'll have to adjust this in DailyDuty once this hits live.

PR #323, opened 2023-02-06
Add levequest methods to QuestManager

Haselnussbomber:
- Added `NumLeveAllowances` field to provide the number of leve allowances the player has.
- Added `IsLevequestComplete(ushort levequestId)` to check if a specific levequest has been completed.
- Added `GetNextLeveAllowancesTimestamp()` which provides the player's next leve allowance timestamp.
   Since it needs to be multiplied by 60 in order to become a unix timestamp, I've made it private and added the following two public helper methods:
   - `GetNextLeveAllowancesUnixTimestamp()` provides a unix timestamp.
   - `GetNextLeveAllowancesDateTime()` provides a local DateTime.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/QuestManager.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/QuestManager.cs
@@ -19,6 +19,7 @@ public unsafe partial struct QuestManager
 	[FieldOffset(0xB70)] public fixed byte BeastReputation[0x10 * 17];
 	[FixedSizeArray<LeveWork>(16)]
 	[FieldOffset(0xC80)] public fixed byte LeveQuests[0x18 * 16];
+    [FieldOffset(0xE00)] public byte NumLeveAllowances;
 
 	[MemberFunction("E8 ?? ?? ?? ?? 66 BA 10 0C")]
     public static partial QuestManager* Instance();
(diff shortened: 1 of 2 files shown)


FFXIVClientStructs/FFXIV/Client/Game/QuestManager.cs
```diff
@@ -19,6 +19,7 @@ public unsafe partial struct QuestManager
 	[FieldOffset(0xB70)] public fixed byte BeastReputation[0x10 * 17];
 	[FixedSizeArray<LeveWork>(16)]
 	[FieldOffset(0xC80)] public fixed byte LeveQuests[0x18 * 16];
+    [FieldOffset(0xE00)] public byte NumLeveAllowances;
```
MidoriKami, line 22: tab error

Haselnussbomber: The whole file, or even repository, is mixed with spaces and tabs, sadly.

MidoriKami: Ah yeah, at first glance it seemed like that was the only line, but it's actually the whole file haha. Not that it matters too much.

Haselnussbomber:
@pohky I've resolved the conflicts caused by the other PR. 🙂

Haselnussbomber:
Updated the offsets in data.yml for 2023.02.03.0000.0000.
Force-pushed twice, because I moved `GetBeastTribeAllowance` up by accident.

PR #329, opened 2023-02-11
Add GetUISaveSegment

MidoriKami:
Adds member function to ConfigModule

sub_140649140

image


MidoriKami:
Realizing now that this is a member of whatever is at 0ff_1319465f8 not ConfigModule

PR #344, opened 2023-02-14
Improve QuestManager

MidoriKami:
The constructor for QuestManager does allocate the last element at 0xEE0 size of 0x8, the next sequential struct is allocated at 0xF10 from QuestManager start.

I've been using the value at 0xEF8 in DailyDuty for quite some time, it's the number of currently accepted levequests.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/QuestManager.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/QuestManager.cs
@@ -4,7 +4,7 @@ using FFXIVClientStructs.FFXIV.Application.Network.WorkDefinitions;
 namespace FFXIVClientStructs.FFXIV.Client.Game;
 
 // idk if this is a manager, but I don't know what else to call it
-[StructLayout(LayoutKind.Explicit, Size = 0xEE8)]
+[StructLayout(LayoutKind.Explicit, Size = 0xF10)]
 public unsafe partial struct QuestManager
 {
 	[FieldOffset(0x10)] public QuestListArray Quest;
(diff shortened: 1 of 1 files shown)


Haselnussbomber:
So, I didn't include it in my PR since it was outside the size. I don't know where QuestManager get's allocated/freed, but it seems the ctor is sub_14075E650, which only goes up to a WORD at 0xEE0 (making it at least 0xEE2 in size).
This and the use of 0x1420E0428 (the location of the variable/QuestManager+0xEF8) makes me think it's actually a global variable in the game instead. Just conveniently located right after the QuestManager?!

While the variable is referenced by functions (called by PacketDispatcher_OnReceivePacket) that have to do with active LeveWorks, there is no reference towards QuestManager in them.

By the way, there is also an unreferenced getter for 0x1420E0428 (our variable in question) at `sub_14075DDF0` (`48 8B 05 ?? ?? ?? ?? C3 CC CC CC CC CC CC CC CC 0F B6 05 ?? ?? ?? ??`). Not sure how safe something like this is.

Instead, I thought of doing something like
```cs
var numActiveLevequests = 0;

foreach (var leveWork in questManager->LeveQuestsSpan)
{
    if (leveWork.LeveId != 0)
    {
        numActiveLevequests++;
    }
}

return numActiveLevequests;
```

After all, I'm really not sure what to make of this. I just wanted to share my thoughts. Up to you guys. 🙂

MidoriKami:
Personally I don't like the idea of calculating something by hand when we have a pre-calculated value available.
You're right in that it does seem to be a global variable.

MidoriKami:
`48 FF 05 ?? ?? ?? ?? EB 0D 66 83 3E 00` Could be use to sig the instance, but I find that approach weird since, we know it's just after the end of the QuestManager struct.
And it's been there for at least 2 major patches.

PR #349, opened 2023-02-15
Improve AgentContentsFinder & AddonContentsFinder

MidoriKami:
This does rename `SelectedContentFinderCondition` to `SelectedDutyId`, as this hasn't been merged into dalamud yet, this is probably fine.


MidoriKami:
I'm considering changing the return of those functions to `nint` since that should be supported now, but I don't know if that'll break things or not...

PR #351, opened 2023-02-16
Create UIInputData.cs

Lynesth:
Welp, there's a first for everything or so they say, so here's me making a PR for CS.
Thanks to @UnknownX7 for the original sig and code that made me want to dig here and @MidoriKami for all the help :)

Let me know what's wrong with the PR (if anything, but surely there will be).
Also, naming stuff was hard, please suggest better ideas.

_Note: I didn't build CS, I just used the struct in a plugin to test it, so there might be things that won't work._

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
@@ -38,7 +38,7 @@ public unsafe partial struct UIInputData
     [FieldOffset(0x4EC)] public int CursorYDelta; // Delta since last frame
 
     // At least this is what it seems to be
-    [FieldOffset(0x4F4)] public byte IsGameWindowFocused;
+    [FieldOffset(0x4F4)] public bool IsGameWindowFocused;
 
     /*
      * All the following keyboard keys states are not triggered if the chat input is active
--- a/FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
@@ -5,6 +5,8 @@ namespace FFXIVClientStructs.FFXIV.Client.UI.Misc;
 [StructLayout(LayoutKind.Explicit, Size = 0xA20)]
 public unsafe partial struct UIInputData
 {
+    public static UIInputData* Instance() => Framework.Instance()->UIModule->GetUIInputData();
+
     /*
      * UIFiltered means those are not set if 
      * - the game window is focused and
--- a/FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
@@ -70,7 +70,14 @@ public unsafe partial struct UIInputData
 
     //[FieldOffset(0x77C)] public byte UnkFlag;
     [FieldOffset(0x77D)] public byte KeyHeldKeycode;
-    [FieldOffset(0x788)] public char LastKeyChar; // (actual character made by key combination, not reset after key is pressed)
+
+    /*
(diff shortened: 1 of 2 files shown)


Lynesth:
Other notes:

- I wanted to use VirtualKey directly in the method, but it's in Dalamud so I revert back to int.
- The only 2 keys (on my keyboard at least, if anyone feels like trying to check, please lemme know) where the offsets don't match the VirtualKey values are Num lock and Scroll lock. Hence the specific handling.
- I could also create a new enum with all the keys if that's preferred, I just thought using `(int) VirtualKey` was probably the most dev friendly in the end.

FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
```diff
@@ -0,0 +1,107 @@
...
+    [FieldOffset(0x4D0)] public int MouseWheel; // -1 for scroll down, 1 for scroll up
+    [FieldOffset(0x4D4)] public MouseButtonFlags MouseButtonHeldFlags;
+    [FieldOffset(0x4D8)] public MouseButtonFlags MouseButtonPressedFlags;
+    [FieldOffset(0x4E0)] public MouseButtonFlags MouseButtonReleasedFlags;
+    [FieldOffset(0x4E4)] public MouseButtonFlags MouseButtonHeldThrottledFlags;
+    [FieldOffset(0x4E8)] public int CursorXDelta; // Delta since last frame
+    [FieldOffset(0x4EC)] public int CursorYDelta; // Delta since last frame
+
+    // At least this is what it seems to be
+    [FieldOffset(0x4F4)] public byte IsGameWindowFocused;
```
Caraxi, line 41: This should be a bool

Lynesth: Right, Kami told me, forgot to change it. Thanks.

Caraxi:
Why are the button flags split?

Lynesth:
> Why are the button flags split?

Because I'm dumb :D

FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
```diff
@@ -0,0 +1,160 @@
...
+            offset = 0x1E0;
+        else if (key == 145) // VirtualKey.SCROLL
+            offset = 0x1E4;
+        else
+            offset = key - 8;
+
+        if (offset < 0 || offset > 133)
+            throw new ArgumentOutOfRangeException();
+
+        return (KeyStateFlags)KeyState[offset];
```
JeffyW, line 101: FYI, since you added [FixedSizeArray<KeyStateFlags>] on the field, you could also just do:
return KeyStateSpan[offset];

It's not substantially different here, just want you to be aware :-)

FFXIVClientStructs/FFXIV/Client/UI/Misc/UIInputData.cs
```diff
@@ -0,0 +1,160 @@
...
+    /*
+     * This one is weird. Seems to fire as long as the last key pressed is still held
+     * Except modifiers (ctrl, shift, alt) where it only fires once
+     * If fires less often than the "Down" flag in the array below but more often than the "Held" one
+     * So Im not sure what to make of this or if this is even useful
+     */
+    //[FieldOffset(0x4F8)] public int IsLastKeyboardKeyDownThrottled;
+
+    [FixedSizeArray<KeyStateFlags>(133)]
+    [FieldOffset(0x51C)] public fixed int KeyState[0x4 * 133];
```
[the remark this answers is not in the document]
aers: This actually should throw a diagnostic and not even build.

https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs.InteropSourceGenerators/FixedSizeArrayGenerator.cs#L84

pohky:
the UIInputData struct should be moved into the UI namespace (`Client::UI::UIInputData`)

PR #355, opened 2023-02-16
Even More MJIManager Findings

KazWolfe:
This pull request aims to add Roaming Minion information to MJIPastureHandler.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -570,7 +570,7 @@ classes:
       0x140B71910: LoadDemandResearchPacket
       0x140B71E80: IsRecipeUnlocked # (this, recipeId)
       0x140B71EB0: IsPouchItemLocked # (this, pouchItemId)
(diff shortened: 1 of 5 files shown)


ida/data.yml
```diff
@@ -580,6 +580,12 @@ classes:
   Client::Game::MJI::IslandState:
     funcs:
       0x140B74FD0: ctor
+  Client::Game::MJI::PastureState:
```
KazWolfe, line 583: Should be `PastureHandler`, my bad.

KazWolfe:
Couple minor notes:

- I'm still not 100% sure on `IslandState.CanEditIsland`. It tracks with what I've seen so far and Lua does a lot of things that only would matter if this was set, but I can't say for sure that this is what this field controls.
- I messed up on `IsFunctionLocked`, it really should be `IsFunctionUnlocked`. I can add the old name and an `[Obsolete]` if that's desired for compatibility.

PR #372, opened 2023-03-02
Add Atk functions

MidoriKami:


FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
```diff
@@ -78,6 +78,9 @@ public unsafe partial struct AtkTextNode : ICreatable
     [MemberFunction("E8 ?? ?? ?? ?? 8D 4E 32")]
     public partial void SetText(byte* str);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 80 38 00 74 27")]
+    public partial Utf8String GetString();
```
Haselnussbomber, line 82: This returns a `byte*` (specifically the StringPtr of the Utf8String at `a1 + 0xB8`).

Haselnussbomber (changes requested):
By the way, can you add a `[GenerateCStrOverloads]` to SetText above while you're at it? Thanks! 😄

FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
MidoriKami: Can't use [GenerateCStrOverloads] for `byte*` returns, you use that for functions that take `byte*` as input arguments

MidoriKami:
I have removed the `SetText` functions in AtkTextNode, I only removed the ones that should be autogenerated with `[GenerateCStrOverloads]`

FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
Haselnussbomber: I know. I've edited the comment (to specify SetText), because I didn't understand GitHubs UI, lol.

PR #374, opened 2023-03-02
More PlayerState research

Haselnussbomber:
Some notes:

The length of `PSNOnlineID` was taken from Sapphire's [ServerZoneDef.h](https://github.com/SapphireServer/Sapphire/blob/ffdbed7/src/common/Network/PacketDef/Zone/ServerZoneDef.h#L908). Apparently the ID is a maximum of 16 characters long, so I guess the 17th byte is for the null terminator.

`PlayerState.PenaltyTimestamps` is used by `RouletteController.GetPenaltyRemainingInMinutes`, which I also have added. It's used by `AgentContentsFinder`, so I added the fields there too. Don't know what the second penalty timestamp is used for.

`FishingBait` was at the wrong offset.

Also I couldn't figure out index 1 for `SelectedPoses` or what `NoviceNetworkFlag.Unknown14` checks. Maybe someone else knows more.

`NumOwnedMounts` is incremented at `0x14090DD5F` while reading in the mount unlock flags. Not sure if this gets updated after acquiring a new mount later on.

Names for `DeliveryLevel`, `IsMeisterFlag`, `IsMeisterFlagMaxCount`, `NoviceNetworkFlag.IsLoginSecurityToken` and `NoviceNetworkFlag.IsBuddyInStable` taken from lua functions. /shrug

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
@@ -83,9 +83,9 @@ public unsafe partial struct PlayerState
     /// 6 = Accessorie Pose: Glasses, Wings
     /// </remarks>
     [FieldOffset(0x47A)] public fixed byte SelectedPoses[7];
-    [FieldOffset(0x481)] public byte NoviceNetworkFlags1;
-    [FieldOffset(0x482)] public byte NoviceNetworkFlags2;
-    [FieldOffset(0x483)] public byte NoviceNetworkFlags3;
+    [FieldOffset(0x481)] public byte PlayerStateFlags1;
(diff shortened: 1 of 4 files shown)


pohky:
not sure if NoviceNetworkFlag is the right name if it has flags for things unrelated to the NoviceNetwork (IsBuddyInStable, IsLoginSecurityToken)
it's also defined as a 3 byte array (looks more like 3 separate fields than an array?) but the flags are uint

Haselnussbomber:
Open for name suggestions. 😛
Any yes, you're right, they are 3 separate bytes. Wasn't sure if i should add them separately. But it's definitely not a uint/dword, since the offset right after the third byte (+484) is used in `Client::Game::UI::PlayerState_IsSecretRecipeBookUnlocked`.

PR #376, opened 2023-03-03
Add Grand Company Structs

MidoriKami:
These structs are already in data.yml

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentGcArmyExpedition.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentGcArmyExpedition.cs
@@ -20,8 +20,8 @@ public unsafe struct GcArmyExpeditionData
 {
     [FieldOffset(0x10)] public int NumEntries;
     
-    [FixedSizeArray<MissionInfo>(18)]
-    [FieldOffset(0x18)] public fixed byte MissionInfoArray[0x78 * 18];
+    [FixedSizeArray<MissionInfo>(34)]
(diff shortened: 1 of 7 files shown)


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentGcArmyExpedition.cs
```diff
@@ -0,0 +1,31 @@
...
+    
+    [FieldOffset(0x40)] public int SelectedTab;
+    [FieldOffset(0x44)] public int SelectedRow;
+}
+
+[StructLayout(LayoutKind.Explicit, Size = 0x920)]
+public unsafe struct GcArmyExpeditionData
+{
+    [FieldOffset(0x10)] public int NumEntries;
+    [FieldOffset(0x18)] public MissionInfo* MissionInfoArray;
```
pohky, line 22: i think this should be a FixedSizeArray instead of a pointer?

MidoriKami: I don't know what it's actual size is, I suspect it's variable, because it includes all the items in your armory aswell, but for me there wasn't dozens of blank spaces after the struct.

MidoriKami: It's strange at least in the memory it looks like it has pre-allocated several more entries.
Only the first 18 are used, so I'll mark it as a fixedsizearray(18)

PR #395, opened 2023-03-07
Offset fixes and cleanup

JeffyW:
In attempting to export structs to headers, I found a bunch of offsets that were off.  Here are fixes.

Also, leveraging Obsolete attribute as a way to mark redundant or legacy properties and suggest the current approach (then excluding Obsolete fields from the exported values).


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentFreeCompanyProfile.cs
```diff
@@ -45,7 +45,7 @@ public unsafe partial struct AgentFreeCompanyProfile
         [FieldOffset(0x27A)] public byte Unk27A;
         [FieldOffset(0x27B)] public byte Unk27B;
 
-        [StructLayout(LayoutKind.Explicit, Size = 0x7)]
+        [StructLayout(LayoutKind.Explicit, Size = 0x6)]
```
pohky, line 48: this doesn't seem right if there is a Recruitment byte at offset 6 it needs to be at least 7 big

JeffyW: Yeah, something's not right.  The part that I was hitting was simply:
        [FieldOffset(0x064)] public FCProfile Profile; // size 0x7 means next bit is 0x6B...
        [FieldOffset(0x06A)] public GrandCompany GrandCompany;
It looks Recruitment is actually at offset 0x5 not 0x6.  Updated.

JeffyW: @pohky let me know if there's anything else you see a problem with.

wolfcomp:
Most of these have already been committed to CS in other PRs

PR #400, opened 2023-03-10
Fix InfoProxyCommonList size offset

Stary2001:
CharacterArray isn't actually a fixed size 200 list, it's variable based on the size value (which is how many entry structs are present, not bytes). I'm not sure how to handle that so I'm just leaving it for now.

Verified with ghidra on InfoProxyCommonList ctor:


Koenari:
Yeah, both Array and Dict a are generally just pointers to an array. With current CS I didn't find a better way to do these and have them typed as well. I should've wrote a comment there.
And yeah Size was a terribly inaccurate name, if you allow it, I'll give a review with a good name for the four size related fields later today

pohky:
#401

PR #402, opened 2023-03-13
add character model id

illion20:


illion20:
@pohky I noticed this didn't get added in the last batch, any issue with the naming or is it something else? Appreciate the feedback, thank you.

pohky:
i tried to look at what this ID is but i didn't really find anything relating to any models
what is it actually supposed to be or what is it used for?

illion20:
For instance in TOP Omega-F can have the "blade legs" model and the "normal" caster model and this byte is set to 4 and 0 respectively. My observation is that when you receive a NPCSpawn packet there is one byte within that packet which determines the graphical model. For the above example only one byte is different for those two actors. Naturally I call this ModelID but any other name is fine too.

PR #404, opened 2023-03-13
Add GetTeleportCost and describe formula.

Ottermandias:
Let me know if this should be done differently.


Caraxi:
I believe this is a static function, so will need to be declared as one in client structs otherwise it'll use `Telepo` instance as a1

Ottermandias:
Oh yeah, I actively looked into where or how to put free functions and then forgot to actually write the static, derp.

PR #423, opened 2023-04-12
Fix MoveItemSlot parameter types

marimelon:


Caraxi:
These should be unsigned shorts

PR #427, opened 2023-05-01
CExporter fixes

awgil:
This is a first bunch of changes I've made to improve the headers generated by CExporter. The goal is to simplify migration to new fresh idbs after patches. So far, here's what I've done:
* started with regenerating headers without making any changes; this creates a base that is easy to compare further changes against
* fixed most of the errors/warnings (easy ones) that appear while running exporter
** exporter now skips code-generated substructs
** added IDAIgnore attribute, which is used mostly for overlapping fields (e.g. matrix types have both multiple component fields and a single data array, 'union' logic of the exporter can't really handle that; while the logic could be improved, I feel that automatic hexrays output is significantly worse in the presence of unions)
* improved exporter handling of arrays (create actual arrays of correct types rather than raw byte arrays)
* fixed ordering of structures in exporter output: do not emit structure X immediately when encountering X* field, since X might depend on not-yet-defined types - forward declaration suffices in such cases
* made vtables (both structs and pointer fields) adhere to hexrays naming convention
** additionally made a pass over existing structures to standardize vtable field naming convention to 'VTable', since that is what generator uses; this removed majority of bogus anonymous unions in exporter output
** additionally made generator emit void* for all vtable members, including gaps
* added support for inheritance that matches what hexrays expects
** custom attribute to manually control whether field is a field or a base class
** default heuristic - structure at offset 0 with name matching type name is a base class
* fixed few obvious bugs (wrong sizes/offsets, incorrect union size calculation, incorrect gap filling before unions, added some additional validation)

There are more things I'd like to do if this PR is accepted:
* improve handling of pointer template arguments for standard containers (currently Pointer<> wrapper is not handled correctly by the exporter)
* the most important - generate extra output containing function signatures, which can then be later imported into database by a dedicated python script (either idarename or a separate one)
* create structures for globals, showing any overlapping names and where they map to in created structure (this would be done by a python script as part of signature propagation flow)
* improve exdrename flow to generate structures for all excel sheets and automatically assign signatures to accessors found by exdrename


NotNite:
I've got a (simpler) PR that gets the basics of CExporter working again - https://github.com/aers/FFXIVClientStructs/pull/468. Only noticed this work *after* I hacked it to working again, sorry!

The only improvement I can see from my PR that could go here is that some type names (particularly `EventHandler`) don't like being imported because they conflict with base IDA types (from my understanding). My hacky fix was to add an underscore, but that breaks idarename. Perhaps you have a better idea?

wolfcomp:
Major changes have happened to CExporter as of #480 and most issues have been fixed with subsequent prs from that point onwards

PR #431, opened 2023-05-15
add UIState.Inspect

Haselnussbomber:
Couldn't figure out ItemIds starting at 0x90, so I left them out.


Haselnussbomber:
Rebased and offset in UIState.cs and address for AddonBuddyInspect updated for 6.4.

PR #432, opened 2023-05-15
Update AtkArrayData

Haselnussbomber:
I noticed that `HasModifiedData`, which I changed in 48bc59e9, can actually have a value of 2, so it's not a bool.

Investigation revealed that
- if set to 0, nothing happens,
- if set to 1, `OnRequestedUpdate` is called on the subscribed addons, but only if the flag checks in `AtkUnitManager.UpdateAddonByID` pass,
- if set to 2, it forces an `OnRequestedUpdate` call on the subscribed addons, bypassing the flag checks.

I renamed the field to `UpdateState` and changed it back to `byte`.

Other than that, I found out:
- Some Addons subscribe to an array in OnSetup and unsubscribe in the Finalize functions, I added the called functions as `SubscribeAtkArrayData`/`UnsubscribeAtkArrayData`. The ID of the addon is stored in `fixed byte SubscribedAddons[16]` at 0xC (length actually just a guess, but I found that StringArray#5 is subscribed by 15 addons).
- `Unk1C` is still unknown and I didn't check it, maybe a 17th addon.
- `Unk1D` is now `SubscribedAddonsCount`.
- `Unk1F` is now `RefCount` as it stores a count of how many Agents are using this array. This is incremented with `IncRefNumberArrayData`/`IncRefStringArrayData` in Show functions, or decremented with `DecRefNumberArrayData`/`DecRefStringArrayData` in Hide functions.

If `RefCount` is -1, the IncRef* functions will set it to 1. When decrementing the ref or unsubscribing from an array, it checks if `RefCount` and `SubscribedAddonsCount` are 0. If this is true, the array will self-destruct: it sets everything to zero, frees memory and removes itself from the AtkArrayDataHolder. Could be dangerous. Maybe just add these functions to data.yml, but not to the .cs files? I doubt anyone would need to call them anyway.

Also, I added the old field names with an Obsolete attribute for compatibility.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
@@ -31,6 +31,18 @@ public unsafe partial struct RaptureAtkModule
     [MemberFunction("E8 ?? ?? ?? ?? 0F B6 44 24 ?? 48 89 9F")]
     public partial bool ChangeUiMode(uint uiMode);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 48 39 77 28 0F 84")]
+    public partial bool IncRefNumberArrayData(int index);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 8B 75 28")]
+    public partial bool DecRefNumberArrayData(int index);
+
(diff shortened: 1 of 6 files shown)


Haselnussbomber:
Rebased and updated for 6.4.

Haselnussbomber:
Rebased and updated for 6.45.

PR #442, opened 2023-05-24
Updating Character StatusFlags

Tischel:
This is incomplete / maybe wrong.
I can't figure it out entirely but decided to at least show my findings with the hopes that someone smarter can improve it.


Aireil:
These should probably have properties associated to them, because people are always confused on how to use StatusFlag and it would avoid this kind of issue where the flags aren't in the same byte anymore. 🤔

PR #451, opened 2023-05-26
Update AgentContentsFinder

MidoriKami:
DailyDuty uses this value to re-sync reward counts for the Raids Tracker


MidoriKami:
It's possible that there are other parts of the agent that moved 0x20 bytes, these were the only values I could verify

PR #454, opened 2023-05-26
Add AgentGettersGenerator

Haselnussbomber:
This Pull Request introduces a generator that enhances the agent structs (with AgentAttribute on them) and AgentModule by adding an `Instance()` getter to the agents and a corresponding `GetAgent*()` getter to AgentModule.

The purpose of this enhancement is to address some confusion that arises when certain agents lack the Instance() getter, while others lack a getter in the AgentModule. Additionally, it aims to reduce the likelihood of crashes (e.g. caused by calling functions on the wrong agent) and the need to recompile plugins after a patch.

As probably known, the C# compiler transforms the reference to a value of the AgentId enum to the plain uint value when compiling a plugin. By adding this generator, plugin developers can avoid using `GetAgentByInternalId(AgentId.*)`, eliminating the need to recompile their plugins solely because the AgentId enum changes after a patch.

Some necessary adjustments had to be made:
- [Breaking Change] Moved AgentHudLayout to the correct namespace (`FFXIVClientStructs.FFXIV.Component.GUI` -> `FFXIVClientStructs.FFXIV.Client.UI.Agent`)
- Removed existing Instance() getter from Agents, which will now be generated
- Removed all exising GetAgent*() getters from AgentModule, which will now be generated
- Added the necessary `unsafe` and `partial` modifiers to every agent struct

It is open for discussion whether we should utilize both Instance() getters and GetAgent*() getters, or if we should opt for one over the other.

---

Additionally. although not part of this Pull Request, I'd like to propose that all known agents will be added as structs, even if they are currently empty (apart from AgentInterface). This would be helpful because some plugin developers may still use GetAgentByInternalId() to resolve an agent's address.

(Yes, I used ChatGPT to write this PR message. :joy:)


Caraxi:
This introduces breaking changes so I'll hold off on merging it

Haselnussbomber:
@wolfcomp I just rebased, the test workflow is failing.

PR #455, opened 2023-05-26
Fixing Scholar Gauge offsets

Tischel:


MidoriKami:
Dismissed Faerie might be a bool? We didn't used to have the ability to properly represent bools, but we do now.

PR #464, opened 2023-05-29
Update UIState.cs

Haselnussbomber:
Updated offsets for Cabinet and RecipeNote.


Haselnussbomber:
HOLD

Haselnussbomber:
It's not. All good.

PR #468, opened 2023-06-02
Get CExporter actually loading in my IDA

NotNite:
This is a very hacky way of doing it, but the file imports with 0 errors now. Not very confident that this is working well, so I'd like a Ghidra user/someone smarter than me to look it over (that's why it's a draft).


wolfcomp:
Major changes have happened to CExporter as of https://github.com/aers/FFXIVClientStructs/pull/480 and most issues have been fixed with subsequent prs from that point onwards

PR #469, opened 2023-06-02
idarename: auto apply type on instances when available

NotNite:
Only implemented in IDA right now. If you run CExporter/the types are in your IDB, it will set instances to their respective types.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -597,9 +597,9 @@ classes:
       0x140B6F8B0: ctor
       0x140B6FB30: dtor
       0x140B6E570: LoadDemandResearchPacket
-      0x140B6EAF0: IsRecipeUnlocked # (this, recipeId)
-      0x140B6EB20: IsPouchItemLocked # (this, pouchItemId)
-      0x140B6CDF0: IsFunctionUnlocked # (this, functionId)
+      0x140B6EAF0: IsRecipeUnlocked(Client::Game::MJI::MJIManager* this, _ recipeId) -> bool
+      0x140B6EB20: IsPouchItemLocked(Client::Game::MJI::MJIManager* this, _ pouchItemId) -> bool
+      0x140B6CDF0: IsFunctionUnlocked(Client::Game::MJI::MJIManager* this, _ functionId) -> bool
       0x140B6D020: HandleActorControlPacket
       0x140B6DB50: HandleInboundPacket
       0x140B6EFA0: GetVisibleMinimapIcons
--- a/ida/ffxiv_idarename.py
+++ b/ida/ffxiv_idarename.py
@@ -189,6 +189,16 @@ class BaseApi(object):
         :rtype: Optional[str]
         """
 
+    @abstractmethod
+    def try_set_type(self, ea, type):
+        """
+        Attempt to set a type on an address.
+        :param ea: Address
+        :type ea: int
+        :param type: Type name
+        :type type: str
+        """
+
 
 api = None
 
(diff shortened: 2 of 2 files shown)


NotNite:
Working on getting this applying types to vfuncs and funcs too - is it a safe assumption that all `func`s in data.yml's a1 will be `type*`? I'll push what I have soon - it checks a1 for being a pointer/int64 - but I want to make sure.

NotNite:
Proper type definitions in data.yml is planned in The Future:tm: with this - maybe we can add support for typing functions in it with this? Right now, at least for MJIManager, @KazWolfe did the following:

```yml
Client::Game::MJI::MJIManager:  # Island Sanctuary
  instances:
    - ea: 0x14218B020
      pointer: True
  funcs:
    0x140B6C400: Initialize # static
    0x140B6C450: Destroy    # static
    0x140B6C490: GetSingleton  # static
    0x140B6F8B0: ctor
    0x140B6FB30: dtor
    0x140B6E570: LoadDemandResearchPacket
    0x140B6EAF0: IsRecipeUnlocked # (this, recipeId)
    0x140B6EB20: IsPouchItemLocked # (this, pouchItemId)
    0x140B6CDF0: IsFunctionUnlocked # (this, functionId)
    0x140B6D020: HandleActorControlPacket
    0x140B6DB50: HandleInboundPacket
    0x140B6EFA0: GetVisibleMinimapIcons
    0x140B6EB30: IsItemLocked
    0x140B6ECF0: GetPastureSlotCount
    0x140B6EE80: GetFarmSlotCount
```

I propose either parsing the comments or allowing you do something like `0x140B6CDF0: IsFunctionUnlocked(Client::Game::MJI::MJIManager* this, uint8 functionId)` and it'll parse the function arguments. If this is a design we want, I can look into implementing it.

KazWolfe:
I really like the premise of this idea, but it raises a couple concerns:

1. This would only be useful so long as our offsets/structs/everything are correct. A naïve readthrough of an IDB that has had this applied to could cause notable confusion or annoyance, especially during post-update periods or with systems that we don't regularly verify for offset correctness. A dev exploring those systems with invalid type hints can very easily be led astray.
2. IDA is not particularly good at determining how many arguments something has. It _generally_ gets things right, but it also definitely can get confused. How would this interact with cases where arguments aren't solidified? Or, worse, arguments change?

I'm not particularly enthusiastic about the idea of using comments for things, as there's no structure for comments nor do I think anyone bothers checking them for correctness. If we want to add a system like this, we should give it a proper set of fields/similar unless we don't really have a choice in the manner.

NotNite:
I agree with your concerns, don't have much commentary to add to them, but one more problem I've noticed: my IDB has tripled (!!!) in size, from 500MB to 1.5GB.

NotNite:
^ That latest commit adds a proof of concept system for declaring argument types in data.yml. You can do this:
```yml
0x140B6EAF0: IsRecipeUnlocked(Client::Game::MJI::MJIManager* this, _ recipeId) -> bool
0x140B6EB20: IsPouchItemLocked(Client::Game::MJI::MJIManager* this, _ pouchItemId) -> bool
0x140B6CDF0: IsFunctionUnlocked(Client::Game::MJI::MJIManager* this, _ functionId) -> bool
```

The arguments (parenthesis and its contents) and the return type (arrow and its following content) are optional. Supply `_` as a type to not replace it and use the current value. When replacing a type bigger than its original value, it is ignored to not cause Decompiler Chaos.

I would like to have a consistent list of types (`byte`, `short`, `int`, `long`, `float`, `double`, unsigned variants) so they can work between Ghidra and IDA, and I also want to deduplicate the code for return type and argument type (notably type parsing and pointer calculation).

wolfcomp:
Great idea in concept, but does not fit the structure of data.yml and could lead to issues if functions parameter calls changes

PR #475, opened 2023-06-08
Make exd structs populate IDA and set functions to return these structs

wolfcomp:


wolfcomp:
This requires the installation of the pip package `requests` to work, as HTTP requests are done through this package.

PR #494, opened 2023-06-25
Update AtkTexture

Haselnussbomber:
First things first, some basic commits:

- aae00be: `UnkBool_2` is actually `bool TextureReady`, but has been marked as obsolete in favor of `IsTextureReady()`, which reads and sets it.
- b7561a1: Adds `DefaultTextureVersion` to AtkModule, which is set as `UiAssetType + 1`, resulting in 1 being Standard resolution, 2 being High resolution.
- ee5119c: Adds `Instance()` getters to UiModule and RaptureAtkModule, because I use these a lot and getting them manually through Framework is getting annoying.

Here's where the fun begins:

- 94526df: I added `LoadTexture` overloads to automatically select the texture version based on the UI Resolution setting (via `DefaultTextureVersion` above), when omitting the version parameter. This originated from a function at 0x1404FEC10. Sadly I can't add a base `MemberFunction` because of `Address` name conflicts, so I added them manually.
- b716a99: Same thing for `LoadIconTexture`.

As you can see, I'm removing the default value for the version parameter in order for the overload to work.
I'm not sure if this is a breaking change because of it.

Random side note: I've seen SimpleTweaks to call `LoadIconTexture` with version 0, which is also valid, since technically anything not 1 gets a `_hr1` slapped onto it.


Haselnussbomber:
- 4b30519: Using CStrOverloads instead. It looks a bit weird, but seems to work. Can undo if wanted.

Haselnussbomber:
- 628fe57: Removed the Obsolete attribute from the new `TextureReady` field. (I force-pushed to add a comment instead.)

Haselnussbomber:
Hold!
I'm using this already and just crashed inside AtkImageNodeRenderer. Not sure if this is related, so I'll have to investigate.

Haselnussbomber:
Okay, this probably isn't going to work, and not even because of the crashes.

I wrote myself a simple texture explorer and came to the conclusion that the `LoadTexture` function with automatic version detection will not work with paths that already have `_hr1` in their name or textures that don't even have a `_hr1` version.

Technically, we would have to have a wrapper around it that checks if the `_hr1` file exists, and then explicitly call `LoadTexture` with the version we can load:
```cs
private void LoadTexture()
{
    var version = 1;
    var pathHr1 = Path;

    if (!Path.EndsWith("_hr1.tex"))
        pathHr1 = pathHr1.Insert(pathHr1.LastIndexOf('.'), "_hr1");

    if (Service.Data.GameData.FileExists(pathHr1))
        version = RaptureAtkModule.Instance()->AtkModule.DefaultTextureVersion;

    PluginLog.Verbose($"Loading texture {Path} version {version}");
    Texture->LoadTexture(Path, version);
}
```

This renders the overloads useless and also depends on Dalamud/Lumina.

I'll probably change the PR to the first 3 commits, but not now since it's 7 am here and I need to sleep. 😴

Haselnussbomber:
I've removed the `LoadTexture` and `LoadIconTexture` overloads. Sorry for the mess.

PR #498, opened 2023-07-02
UI::Map Add New Data

MidoriKami:
So yeah, that's a thing.

So these data structures contain all the necessary information to display the quests on the map.

Generally speaking there are two kinds of structures uses here, one that contains 4 integer values, and another that contains various other data including a pointer to the parent's tooltip.

These all work, as I am currently using them in Mappy, if any of yall wanna take a crack at simplifying this be my guest.
I have no time urgency to merge this as I have my own copy local to my plugin Mappy.

If you have any better ideas for the names be my guest.


MidoriKami:
Closing for now, I have a strong hunch that each marker category has two sets of structs, one that contains basic marker data, and another that contains some more information such as tooltips.

PR #499, opened 2023-07-02
Add Default Args to UIModule.PlaySound

MidoriKami:


Caraxi:
This isn't a breaking change

wolfcomp:
wasnt sure if it was or not so just safe guessed it

PR #504, opened 2023-07-08
Use FixedSizeArray in RaptureMacroModule

Haselnussbomber:


MidoriKami:
Could fairly easily modify this to not be breaking and mark the properties as obsolete, then remove the obsolete fields with 6.5

This would give devs that need to review their code for Dalamud Api9 a heads up that things will change.

Haselnussbomber:
Sure, that would require different field names though, which I didn't really like.

I could change it to `IndividualMacros` and `SharedMacros`, but what do I do about:
> The type 'RaptureMacroModule.Macro' already contains a definition for 'Lines'
🤔

Haselnussbomber:
I think that sould do it.

MidoriKami:
Very nice! Thank you

Haselnussbomber:
Looking back, I wonder if changing the Macro struct to LayoutKind.Explicit and adding fields is a breaking change. 🤔

wolfcomp:
I would say just remove the Fields and Structs that has Obsoletes in this PR since removing them with 6.5 any way to clean up the structs.

PR #505, opened 2023-07-10
Add reaper shroud container.

Ottermandias:


Ottermandias:
Rebased and updated addresses.

PR #527, opened 2023-07-27
Add .editorconfig

Haselnussbomber:
I'm not sure if this will help to get some consistency in here, but it's worth a try.
This obviously needs a `dotnet format` after merging.

In my opinion, .NET already has good defaults, which I have tweaked a bit. However, I'm not sure they're actually applied.
For example, according to the [docs](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options) `dotnet_sort_system_directives_first` defaults to true, but it didn't sort them correctly unless specified in the .editorconfig file. 🤷‍♂️

Also, since nobody runs `dotnet format` before comitting their changes, we might need a GitHub Action for this. Something similar to the [v9 rollup](https://github.com/goatcorp/Dalamud/blob/master/.github/workflows/rollup.yml) action in Dalamud, which automatically pushes format fixes via a PR that it creates.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,53 @@
+root = true
+
+# All files
+[*]
+end_of_line = crlf
+indent_style = space
+charset = utf-8
+
+# XML project files
+[*.{csproj,vcxproj,vcxproj.filters,proj,projitems,shproj}]
+indent_size = 2
+
+# XML config files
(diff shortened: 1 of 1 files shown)


Haselnussbomber:
Added K&R style rules.

wolfcomp:
Those causes the least amount of write changes on the files that are in the project

wolfcomp:
Spoke with aers and said it looks good.

PR #531, opened 2023-08-07
Add RaptureHotbarModule->SaveSlot

kiwikahawai:
Function copies the provided HotBarSlot to the user's saved hotbars and marks the change for saving in the data files.


kiwikahawai:
Also Kaz's related PR #532. There might be a conflict between the two PRs as we're both appending to the bottom of `RaptureHotbarModule` now, I'm happy to rebase if needed

PR #535, opened 2023-08-08
Partially revert pre-6.48 hotbar changes

KazWolfe:
- Remove Span references, as they don't allow (easily) accessing pointers, which are necessary for certain methods.
- Add GetSlotById for completeness' sake


KazWolfe:
As-is, this PR will completely rollback the Obsoletes and swap to `FixedSizeArray`. If we still want those (and the simpler API surface for saved hotbars), I might be able to add in a helper method `GetHotbarSlot(uint slotId)` to return a pointer

PR #547, opened 2023-08-26
Implement AgentRetainerTask

WorkingRobot:
Unsure about the agent size, I just made my best guess. Let me know if it needs to be changed.


aers:
For future reference, size of a class like this is easily checked because they get heap allocated.

QIyFMUT

And you did get it correct.

pohky:
if you look at the ctor it's an array of 2 uints
so you can just make it `[FieldOffset(0x50)] public fixed uint RewardItemId[2];`

PR #549, opened 2023-08-27
 Add some graphics structs, fields and functions

Exter-N:
Most of these come from modding-related plugins and/or shader modding research.

Also adds a few helpers, mostly `Span<T>`-based.

cc @Ottermandias @chirpxiv


FFXIVClientStructs/FFXIV/Client/Graphics/Render/Material.cs
```diff
@@ -1,9 +1,44 @@
+using FFXIVClientStructs.FFXIV.Client.Graphics.Kernel;
 using FFXIVClientStructs.FFXIV.Client.System.Resource.Handle;
 
 namespace FFXIVClientStructs.FFXIV.Client.Graphics.Render; 
 
 [StructLayout(LayoutKind.Explicit, Size = 0x20)]
```
wolfcomp, line 6: Find the new size for this, as it overruns the size constraints with the field offsets

Exter-N: Good catch, it is indeed `0x40` + a variable part, fixed it + added a comment.

wolfcomp:
Pull the newest main and reformat your code with the `.editorconfig` preset

PR #551, opened 2023-08-28
Create CExporter action

wolfcomp:
This is a preliminary action will add one more before this should be merged which will push cexporter on new struct changes in CS


wolfcomp:
Both actions now there so good to merge. 

Yes, this will run a workflow on all PRs that modifies CS to stop offsets and what not to show if something seems wrong with sizes. *I'm tired of fixing other people's miss types when changing something in CExporter*

PR #554, opened 2023-08-28
data.yml in solution

Ottermandias:
Add data.yml as Solution Item to the solution so it is immediately loaded and searchable in VS/Rider.
Discard if not wanted.


wolfcomp:
I personally just used VS Code in the folder to edit those things, but I can see this being useful for others.

MidoriKami:
I also would appreciate this being merged.

PR #556, opened 2023-08-30
Patch 6.5 remove all obsoletes in current code

wolfcomp:


wolfcomp:
This should be the last one of the breaking changes to be merged of all the 6.5 changes.

FFXIVClientStructs/FFXIV/Component/GUI/AtkResNode.cs
```diff
@@ -98,8 +97,6 @@ public unsafe partial struct AtkResNode : ICreatable {
 
     // asm accesses these fields together so this is one 32bit field with priority+flags
     [FieldOffset(0x9C)] public ushort Priority;
-    [Obsolete("Use NodeFlags", true)]
-    [FieldOffset(0x9E)] public short Flags;
     [FieldOffset(0x9E)] public NodeFlags NodeFlags;
     [FieldOffset(0xA0)] public uint Flags_2; // bit 1 = has changes, ClipCount is bits 10-17, idk its a mess
```
WorkingRobot, line 101: This field is identical to the one below, you might wanna remove this line, too.

wolfcomp: I haven't pulled in a while for what is on the main branch as I have not been focused on removing these before 6.5 patch.
The fact that these are marked as error with the compilation if used will make it a non-breaking change later when they are removed.

wolfcomp:
This should possibly not be merged before 6.51 to give people time to update all the different plugins.

PR #562, opened 2023-09-04
Update RaptureGearsetModule

Haselnussbomber:
**Additions:**
- Added `int FindGearsetIDByName(Utf8String* gearsetName)`.
- Added `GearsetItemFlag` enum which is used in the new `GearsetItem.Flags` field.
- `Unknown04` in `GearsetFlag` identified as `MainHandMissing`.
- Remarks to known `GearsetFlag` values.
- XML docs to `GetBannerIndex`, `SetBannerIndex` and `HasLinkedBanner` (also renamed them, see below).

**Replacements:**
- Replaced `Gearsets` class and `Gearset` field with a `FixedSizeArray<GearsetEntry>(100)` field called `Entries`.
- Replaced `ItemsData` field with a `FixedSizeArray<GearsetItem>(14)` field called `Items`.
  - I added a helper enum called `GearsetItemIndex` which can be used as an index for the ItemsSpan, for example:  
`var mainHand = gearsetEntry.ItemsSpan[(int)GearsetItemIndex.MainHand];`

**Return type changes:**
- `IsValidGearset` and `HasLinkedGlamourPlate` now return `bool`, because that is really all they can return.
- `CreateGearset` now returns a `sbyte`, where `-1` means it failed.
- `GetBannerIndex` now returns a `sbyte`, where `-1` means the gear set was not linked to a banner.

**Removals:**
- Removed `vtbl` in favour of `RaptureGearsetModule.StaticAddressPointers.VTable` (generated via `VTableAddressAttribute`).
- Removed `ModuleName`, which was obsolete already (=> `UserFileEvent.FileName`).

**Renames:**
These were really long and I got inspred by `LinkGlamourPlate` and `HasLinkedGlamourPlate` to shorten them:
- `GetBannerIndexByGearsetIndex` renamed to `GetBannerIndex`.
- `SetBannerIndexForGearsetIndex` renamed to `SetBannerIndex`.
- `IsGearsetLinkedWithBanner` renamed to `HasLinkedBanner`.

Of course, these changes are intended to be merged when patch 6.5 arrives.
I'm open to feedback, especially on changes to the return types (I don't mind changing them back if it's unwelcome, but this is how I think they're being used). 😉

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureGearsetModule.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureGearsetModule.cs
@@ -101,9 +101,9 @@ public unsafe partial struct RaptureGearsetModule {
     /// Get the Banner index of a Gearset.
     /// </summary>
     /// <param name="gearsetId">The ID of the gearset.</param>
-    /// <returns>The actual BannerIndex (gearset->BannerIndex - 1).</returns>
+    /// <returns>The Banner index, or -1 if it was not linked.</returns>
     [MemberFunction("E8 ?? ?? ?? ?? 0F B6 C0 41 3B C7")]
-    public partial byte GetBannerIndex(byte gearsetId);
+    public partial sbyte GetBannerIndex(byte gearsetId);
 
     /// <summary>
     /// Set the Banner index of a Gearset.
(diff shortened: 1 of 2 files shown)


Haselnussbomber:
Reverted the removal of `GearsetItem` fields in `GearsetEntry` to avoid headaches involving ItemSpan and getting pointers.

PR #576, opened 2023-09-15
Create AtkTextureResourceManager and update AtkTextureResource & ResourceHandle

WorkingRobot:
This PR implements AtkTextureResourceManager (I'm unsure about the naming for right now), adds more vfunc definitions for ResourceHandle, and names fields in AtkTextureResource


Haselnussbomber:
Like ares said in Discord, we are merging a lot of breaking changes next week. This patch will be especially breaking, since we’re kinda doing a cleanup in the ClientStructs code base. Lots of things are moved around, fixed, replaced by FixedSizeArrays, and so on. We are also removing all obsoletes (#556), so I think it’s fine to leave out the properties.
Besides, I was the one that added DefaultTextureVersion 3 months ago and according to GitHub search nobody uses it, so nothing will break anyway. 🤷‍♂️

PR #581, opened 2023-09-15
Update RetainerCommentModule

Haselnussbomber:
- Fixed offset of `Retainers` and fixed `RetainerComment` size.
- Replaced `RetainerCommentList` with `FixedSizeArray<RetainerComment>(10)`.
- Replaced the `Comment` property with a fixed byte array.
- Made `RetainerComment` struct layout explicit.
- Changed return type of `SetComment` from `void*` to `void`.


Haselnussbomber:
You know what? I think I'll do one PR for everything.

PR #582, opened 2023-09-16
FixedSizeArray cleanup

Haselnussbomber:
This PR removes structs that are no longer necessary because of FixedSizeArrays, and some obsolete fields that may have been part of the cleanup in #556.

Not included are:

- `RaptureHotbarModule`, because Kaz left a comment waiting for "better fixed arrays".
- `RaptureGearsetModule` (handled by #562).
- `RaptureMacroModule` (handled by #504).

Also fixed `RetainerCommentModule` and removed the Comment property in favour of a cleaner fixed byte array (since the offset and size was wrong, I assume nobody used it anyway).


Haselnussbomber:
Tests are failing because of dab84b2ff24761e1ab023409bd181f4d8c18678a (not part of this PR).

wolfcomp:
pull newest main and it resolves it

PR #590, opened 2023-09-19
Add some height and emore related stuff.

Ottermandias:


pohky:
hm pretty sure that `CalculateHeight` returns float and `GetHeightAdjustActionTimelineRowId` looks like uint return but not sure
`EmoteController` belongs in the `Control` namespace -> `Client::Game::Control::EmoteController`
the field can then also be named EmoteController i guess

Ottermandias:
Ghidra is very firm in that it returns ulongs in the function and in most of the callsites, but some cast it to float. Float would definitely make more sense.
And yeah, the row id is definitely a uint, technically, but it also doesn't matter, I think?

PR #594, opened 2023-09-24
Add AtkEventDispatcher

Haselnussbomber:
Didn't investigate further and don't plan on it either right now, sorry.


FFXIVClientStructs/FFXIV/Component/GUI/AtkStage.cs
```diff
@@ -17,7 +17,9 @@ public unsafe partial struct AtkStage {
     [FieldOffset(0x78)] public AtkDragDropManager DragDropManager;
     [FieldOffset(0x168)] public AtkTooltipManager TooltipManager;
     [FieldOffset(0x338)] public AtkCursor AtkCursor;
-    [FieldOffset(0x350)] public void* AtkEventDispatcher;
+    [FixedSizeArray<Pointer<AtkEventDispatcher>>(32)]
```
pohky, line 20: i think this was meant to just be a `[FixedSizeArray<AtkEventDispatcher>(32)]` without the Pointer?

Haselnussbomber: Yep, seems to be a bad habit of mine. My ReClass had it right, without Pointer. :zany_face:

PR #599, opened 2023-09-25
More Addons for data.yml

Haselnussbomber:
Added some addons I use, and some I don't use.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -7073,6 +7073,24 @@ classes:
         base: Component::GUI::AtkUnitBase
     funcs:
       0x140D61FD0: ctor
+  Client::UI::AddonBuddy:
+    vtbls:
+      - ea: 0x141A677A8
+        base: Component::GUI::AtkUnitBase
+    funcs:
+      0x140F7DAF0: ctor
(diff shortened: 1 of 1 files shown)


ida/data.yml
```diff
@@ -7290,17 +7308,87 @@ classes:
...
+    funcs:
+      0x140F76F00: ctor
   Client::UI::AddonMonsterNote:
     vtbls:
       - ea: 0x141A526E0
         base: Component::GUI::AtkUnitBase
     funcs:
       0x140EB5810: ctor
-  Client::UI::AddonGatheringNoteBook:
+  Client::UI::AddonGatheringNote:
```
pohky, line 7391: this is `AddonGatheringNoteBook`
addons are usually named after what is used in `LoadUldResourceHandle` or the name from the old RTTI

Haselnussbomber: I renamed this because `/tweaks debug` said it's called `AddonGatheringNote`. 🤷‍♂️

Haselnussbomber: Hmm yeah, classinformer.csv has it as `AddonGatheringNoteBook`. I'll change it back.

PR #601, opened 2023-09-27
QuestLinkMarkers

MidoriKami:


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentMap.cs
```diff
@@ -194,6 +197,27 @@ public struct OpenMapInfo {
...
+public unsafe partial struct QuestLinkContainer {
+    [FieldOffset(0x08)] public ushort MarkerCount;
+
+    [FixedSizeArray<QuestLinkMarker>(20)]
+    [FieldOffset(0x18)] public fixed byte Markers[0x88 * 20];
+}
+
+[StructLayout(LayoutKind.Explicit, Size = 0x88)]
+public struct QuestLinkMarker {
+    [FieldOffset(0x02)] public ushort QuestId;
```
wolfcomp, line 210: you should add `[FieldOffset(0x00)] public bool Valid;` to this struct

MidoriKami: image
That value does not indicate it is valid. As you can see in this image the value is still 1 even though MarkerCount is zero.
I will not be including it in this PR.

wolfcomp: In terms of the game code it checks this bit if it is 0 or 1 when processing packets and therefor should be included

MidoriKami: I am not going to put my name on something that I don't believe in. If you want it added, you can add it in a different PR so it puts your name on it, so people can question you on what it means.

PR #604, opened 2023-10-02
Don't use potential runtime reflection in Resolver

AsgardXIV:
The cache in the resolver currently uses code that potentially requires runtime reflection to serialize/deserialize the cache.

In theory this isn't compatible with trimming or AOT and produces warnings. 
Switching to the source code generation for System.Text.Json resolves this. See more at: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation

With this change, the project builds cleanly with no warnings when trimmed or publishing as part of an AOT project (and I've confirmed it actually functions when built and injected as a native library, at least on initial testing).


aers:
Thanks.

PR #618, opened 2023-10-03
Restore ICreatables

MidoriKami:


wolfcomp:
will probably investigate a better way of doing this if no one beats me to it

PR #638, opened 2023-10-06
Add ffxiv_typedecl

NotNite:
It's just #469 on a separate file.


wolfcomp:
Deemed not necessary due to recent updates to CExporter

PR #645, opened 2023-10-11
Remove Incorrect Map Structs

MidoriKami:
These "SimpleMapMarkerContainers" are actually StdDeques<SimpleMapMarker>
These Deques only have valid data during load, and otherwise have a size of zero. The only reason data still even exists there is because square doesn't clean it up.

I'm opting to remove these structs entirely as they are not particularly useful to start with. All the same data and more is available vis the MapMarkerContainers instead.


MidoriKami:
Github codesearch indicates that no projects whatsoever use the removed fields.
Additionally, use of these fields is improper in the first place as the data only still exists due to square not clearing unused memory.

PR #654, opened 2023-10-18
More PlayerState/UIState findings

Haselnussbomber:
**Fixed:**
- Bitmask field sizes in both PlayerState and UIState were wrong
  - My bad. They were one byte too short using `rowCount >> 3`. Now it uses `(rowCount + 7) >> 3` to round it up (this is equivalent to `Math.Ceiling(rowCount / 8)`).
- Offsets for Aetheryte ids were wrong and I fixed them, but...

**Bugged:**
- ...Aetheryte ids are ushort, not byte. Changing them would be a breaking change, so I opened a separate PR: #655

**Added:**
- `FreeAetherytePlayStationPlus` (not usable in Windows client)
- Fishing unlocks:
  - `CaughtFishBitmask` and `CaughtSpearfishBitmask`, found via GatherBuddy signatures
  - `UnlockedSpearfishingNotebookBitmask`
- Ornament unlocks:
  - `UnlockedOrnamentsBitmask`
- Sightseeing Log unlocks:
  - `UnlockedAdventureBitmask` and `UnlockedAdventureExPhaseBitmask` with corresponding functions `IsAdventureComplete` and `IsAdventureExPhaseComplete`
  - `SightseeingLogUnlockState` and `SightseeingLogUnlockStateEx`
- `IsPublicContentCompleted` and `IsPublicContentUnlocked`
- Unlocked Folklore Tomes:
  - `UnlockedMinerFolkloreTomeBitmask`, `UnlockedBotanistFolkloreTomeBitmask` and `UnlockedFishingFolkloreTomeBitmask` found via the `PlayerState.IsFolkloreTomeUnlocked` function
- `HasNewGcArmyCandidate`

PR #655, opened 2023-10-18
Fix PlayerState aetheryte id field types

Haselnussbomber:
This PR comes after #654.
This is a **breaking change**.

The fields `HomeAetheryteId`, `FavouriteAetheryteArray` and `FreeAetheryteId` are ushort, not byte.


wolfcomp:
Couldn't find uses of these fields

PR #658, opened 2023-10-20
Update AddonRepair for 6.5

marimelon:


pohky:
those last 2 buttons (the left right) are part of another 0x38 sized class at 0xF7B0 (the dropdown) thats also used by other addons
could you reomove those as they don't belong directly into the AddonRepair (or comment them out with a note that they are part of another class)

marimelon:
Deleted.
Thank you for your detailed explanation.

PR #661, opened 2023-10-21
Add GcArmyManager

Haselnussbomber:
Some data about Grand Company Squadrons.


pohky:
Flags 1 is in a mission, Flags 2 is in training
could add those at least just so it's not empty

PR #663, opened 2023-10-21
Update InventoryManager

Haselnussbomber:
Adds the following functions to InventoryManager found via lua functions with the same names:

- `GetEmptySlotsInBag`
- `GetGil`
- `GetCompanySeals`
- `GetMaxCompanySeals`
- `GetTomestoneCount`
- `GetGoldSaucerCoin`

Bonus functions:

- `GetInventorySlot` (shortcut for ->GetInventoryContainer()->GetInventorySlot())
- `GetRetainerGil`
- `GetWolfMarks`
- `GetAlliedSeals`

Possible functions, I can't check:
- `sub_140734260` is probably `uint GetFreeCompanyGil()`
- `sub_14072F8C0` is probably `int SendTradeRequest(uint playerObjectId)`


pohky:
yes those last 2 are what you think they are but not sure if we should add something like sending trade requests, at least as a callable function in InventoryManager

PR #664, opened 2023-10-22
Added a bunch of research around MJI workshop schedules and favors.

awgil:


wolfcomp:
If bytes are 100% certain to be set as either `0x00` or `0x01` you can type the field as `bool` instead of `byte`

awgil:
>If bytes are 100% certain to be set as either 0x00 or 0x01 you can type the field as bool instead of byte

Replaced two instances that I've found.

awgil:
Rebased on master and updated addresses in data.yml

PR #665, opened 2023-10-23
Add AddonConfig

MidoriKami:
Maybe breaking? Because `void*` -> `AddonConfigModule*`


MidoriKami:
Github code search indicates that no Main Repository plugins use this function. 
So while it is `breaking` there are no effected plugins.

PR #673, opened 2023-10-26
Update SatisfactionSupplyManager

Haselnussbomber:
- Margrat was missing
- Fixed offsets
- Added some more stuff

No clue about the struct size since it's part of GameMain, so I didn't touch it.


wolfcomp:
didn't find anyone using anything other than the member functions

PR #677, opened 2023-10-27
Update AgentBannerEditor

Haselnussbomber:
It always bothered me that I didn't reverse these data structures properly, so I had to check them again.
Threw in some obsoletes, so it's not breaking.


wolfcomp:
Do mark Obsolete with `, true` at the end to force users to change over. This is to make it easier to remove them at a later date without having a breaking change suddenly, as it only forces it on compile of the application and not if it is already compiled.

Haselnussbomber:
Right, I forgot. Thanks!

PR #691, opened 2023-11-03
Update ActionBar structs

ItsBexy:
Mapped out the following structs in much more detail: `AddonActionBarBase`, `AddonActionBarX`, `AddonActionBar`, `AddonActionCross`, `AddonActionDoubleCrossBase`. Also identified a number of vfuncs for these addons.

There are still a few stray fields I have been unable to identify, but by and large these structs are nearly complete.

Other notes:
- Renamed the file for `AddonActionCross` for consistency (AddonActionBarCross isn't a thing)
- Renamed a few fields in `AddonActionCross` for clarity (obsoleted old names with warnings)
- Cut down on some of my more verbose summaries
- Bonus: identified a couple functions the game uses to write numbers as special/boxed characters.


ItsBexy:
I've commented out this member function declaration for now; CExporter is not a fan of it, but I'm a little confused because it seems like other member functions in CS are declared similarly without issue. Could the `out` param be throwing a wrench in the works, or did I goof something else up?

```cs
/// <summary>
/// Accepts a value representing the Expanded Hold Controls or WXHB configuration, and converts that value into the corresponding hotbar ID and a bool indicating which half of the bar to reference.
/// </summary>
/// <param name="mapValue">A value from 1-20.<br/>1-16 represent each half of the eight Cross Hotbar sets.<br/>17-20 represent the Cycle Up/Down options.</param>
/// <param name="useLeftSide">True = left side of target bar, False = right side of target bar</param>
/// <returns>The hotbar ID (<see cref="AddonActionBarBase.RaptureHotbarId"/>) of the action set assigned to this input.</returns>
[MemberFunction("E8 ?? ?? ?? ?? 8B D0 83 F8 12 74 38")]
public static partial int GetBarTarget(int mapValue, out bool useLeftSide);
```
```
CS8795: Partial method 'AddonActionCross.GetBarTarget(int, out bool)' must have an implementation part because it has accessibility modifiers.
CS0759: No defining declaration found for implementing declaration of partial method 'AddonActionCross.GetBarTarget(int, bool)'
```
[Here's the build attempt with the errors](https://github.com/aers/FFXIVClientStructs/actions/runs/6748971730/job/18348412632?pr=691)

It's a static function, but I intended to include it with `AddonActionCross` because the client only ever calls it in regard to Cross Hotbars.

pohky:
> **EDIT:** I think I can confirm that using either an `out` or `ref` parameter is in fact the cause of the error.

yes, the generator does not support out/ref arguments but you can use a pointer (instead of `ref bool arg` make it `bool* arg`)

ItsBexy:
Ah, terrific, thank you!

FFXIVClientStructs/FFXIV/Client/UI/AddonActionBarBase.cs
```diff
@@ -5,28 +5,33 @@ namespace FFXIVClientStructs.FFXIV.Client.UI;
 [StructLayout(LayoutKind.Explicit, Size = 0x248)]
 public unsafe partial struct AddonActionBarBase {
     [FieldOffset(0x00)] public AtkUnitBase AtkUnitBase;
-
+    
     [FieldOffset(0x220)] public ActionBarSlot* ActionBarSlots;
```
MidoriKami, line 9: This is most likely a StdVector<ActionBarSlot>

ItsBexy: Yeah, that tracks, I can throw that change in. What new name to give the replacement StdVector, though? The name `ActionBarSlots` is still accurate and clear, it's just the type that would need changing. Hmm...

MidoriKami: `ActionBarSlotVector`

MidoriKami:
image
Seems this line needs to be obsoleted as well, since the vector essentially does what the span did.

PR #693, opened 2023-11-07
Update AtkComponentList, AtkComponentDropDownList

Haselnussbomber:
It is what it is (a mess, as usual).

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentBase.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentBase.cs
@@ -62,6 +62,7 @@ public unsafe partial struct AtkComponentBase {
     [VirtualFunction(5)]
     public partial void OnUldUpdate();
 
+    // TODO: return void
     [VirtualFunction(10)]
     public partial void* SetEnabledState(bool enabled);
 }
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
@@ -7,7 +7,7 @@ namespace FFXIVClientStructs.FFXIV.Component.GUI;
 // common CreateAtkComponent function 8B FA 33 DB E8 ?? ?? ?? ?? 
 // type 1
 [StructLayout(LayoutKind.Explicit, Size = 0x1A8)]
-public unsafe struct AtkComponentList {
+public unsafe partial struct AtkComponentList {
     [FieldOffset(0x0)] public AtkComponentBase AtkComponentBase;
     [FieldOffset(0xC0)] public AtkComponentListItemRenderer* FirstAtkComponentListItemRenderer;
     [FieldOffset(0xC8)] public AtkComponentScrollBar* AtkComponentScrollBarC8;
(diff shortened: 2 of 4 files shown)


FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
```diff
@@ -22,9 +22,50 @@ public unsafe struct AtkComponentList {
...
+    [VirtualFunction(36)]
+    public partial int GetItemCount();
+
+    /// <remarks> Used by <see cref="AtkComponentDropDownList"/>. </remarks>
+    [MemberFunction("E8 ?? ?? ?? ?? 45 38 A4 3E"), GenerateCStrOverloads]
+    public partial void SetItemLabel(int index, byte* text);
+
+    /// <remarks> Used by <see cref="AtkComponentDropDownList"/>. </remarks>
+    [MemberFunction("E8 ?? ?? ?? ?? 89 5C 24 28 48 8B 5C 24")]
+    public partial byte* GetItemLabel(int index);
```
MidoriKami, line 57: Would you be willing to add a helper to get the item label as a string?

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
```diff
@@ -22,9 +22,50 @@ public unsafe struct AtkComponentList {
...
+    public partial void SetItemHighlightedState(int index, bool highlighted, bool triggerUpdate = true);
+
+    [VirtualFunction(29)]
+    public partial bool GetItemHighlightedState(int index);
+
+    [VirtualFunction(31)]
+    public partial void SelectItem(int index, bool dispatchEvent37 = false);
+
+    [VirtualFunction(32)]
+    public partial void DeselectItem();
```
MidoriKami, line 46: In cases like the ContentsFinder, does this deselect the last selected item? all items?

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
Haselnussbomber: I could, but we have `MemoryHelper.ReadStringNullTerminated()`? 🤔

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentBase.cs
```diff
@@ -62,6 +62,7 @@ public unsafe partial struct AtkComponentBase {
     [VirtualFunction(5)]
     public partial void OnUldUpdate();
 
+    // TODO: return void
     [VirtualFunction(10)]
     public partial void* SetEnabledState(bool enabled);
```
MidoriKami, line 67: If this is wrong, shouldn't it get fixed even if it's a "breaking change"?
Returning things when you shouldn't is likely to cause problems.

Haselnussbomber: Dunno. Thats a question for @pohky, I guess.

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
MidoriKami: Normally we can use the attribute to generate string overloads, but we can't do that for a return value, it's just a ease-of-use thing. Up to you.

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
Haselnussbomber: This the duty selection in ContentsFinder is unrelated to the selected item in the list.
See how it changes the SelectedItemIndex when I first click on an item:

https://github.com/aers/FFXIVClientStructs/assets/96642047/e7f31a5e-868a-4a29-ac04-ebff2d54b6d4

When I call DeselectItem, it removes the highlight on the item and sets SelectedItemIndex to -1.

MidoriKami: Ahh I see, so it's just the yellow selection background, which only applies to one item at a time. Thanks for the clarification!

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentList.cs
Haselnussbomber: Hm, nah. :rofl: I don't really like to make a mess here.

Maybe with the next API bump we could think about replacing all byte* with a little helper struct for null terminated strings, so we can still pass them as pointers but also have a C# ToString method. 👀

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentBase.cs
wolfcomp: This is yes wrong and yes breaking, but looking at IDA and how it resolves how to return something based on the return type indicates that it's not major enough until a larger breaking change is required to change this also.

PR #694, opened 2023-11-07
Update RaptureAtkModule

Haselnussbomber:
This makes it possible to hook `OnUpdate` to read out the `AgentUpdateFlag` before it's reset to 0.


pohky:
that function has a float arg, didn't check but probably the delta like the other update funcs

Haselnussbomber:
Yep, you're right. I was wondering about that when I looked in the pseudocode, but didn't check.
deltas

I'll update it.

Haselnussbomber:
Ok, I force-pushed it and also renamed it from `OnUpdate` to just `Update`.

PR #697, opened 2023-11-10
Add AddonJobHud

ItsBexy:
Adds structs for job gauge addons and their associated data.

~~Individual job gauge structs are defined but not yet mapped out. Currently I have them sharing a file, but once mapped they could probably do with being split out into separate files per job.~~

**Update:** Latest commit splits out the addons into their own files per job. A few of them are mapped in detail (melee done so far).


ItsBexy:
- All job gauges are now mapped
- Added some discovered fields/functions for certain types of `AtkComponent`
- Switched to generic names for "Prereq" data fields; these fields are affected by level progression/sync which makes them tricky to identify with confidence.
- Other small adjustments to naming conventions

Some gauges still have some stray unidentified data fields within their addon structs, but all pointer fields are documented, as are the subroutine funcs that they call.

PR #701, opened 2023-11-15
Update AtkComponentGaugeBar

ItsBexy:
PR'ing this separately from the gargantuan JobHud PR, because the changes here are both smaller and more broadly-useful.


FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentGaugeBar.cs
```diff
@@ -6,10 +6,93 @@ namespace FFXIVClientStructs.FFXIV.Component.GUI;
...
+    [FieldOffset(0x188)] public AtkTextNode* ParameterTextNode;         // mainly seems to be used by the Parameter Widget; other gauges that have a text node won't necessarily use this field to point to it
 
     [FieldOffset(0x190)] public int MinValue;
     [FieldOffset(0x194)] public int MaxValue;
+
+    [FieldOffset(0x1A0)] public short MarginX;
+    [FieldOffset(0x1A2)] public short MaxFillPositionX;
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 8B F8 48 85 C0 0F 84 ?? ?? ?? ?? 49 8B 0F")]
+    public partial void Ctor();
```
wolfcomp, line 86: Points to the wrong `ctor` currently is the sig of `140535D30` when it should be `14056F5E0`

PR #707, opened 2023-11-19
Add DynamicEvent (critical engagements)

syldrathecat:
Added DynamicEventManager and DynamicEvent to read the Critical Engagement table inside of Bozja/Zadnor. DynamicEventManager.Instance() returns 0 in any other zone.

While inside of the zone, all 16 entries in the table are fully populated with all data for all possible events (12 Critical Engagements, 3 Duels, and 1 Large Scale Battle) that can spawn there.

I plopped the field in to ContentDirector instead of any of its sub-classes since the DynamicEventManager.Instance() function appears to access the field through what should be `ContentDirector*`.

I also turned it in to an unsafe struct to add my pointer I hope thats ok.


syldrathecat:
image

PR #715, opened 2023-11-29
Add ReadyToDraw to GameObject

AsgardXIV:
Adding some drawing related stuff.

The ReadyToDraw flag is checked when calling `EnableDraw` and it will early exit if the flag is not set.
It's flagged true during an early frame at the top of `Update` on the `GameObject` typically.

Could be argued that it should be called `HasHadFirstUpdate` or `IsReady` similar so if you have a preference for a different name, let me know.

`ObjectTargetableFlags` may need renaming as it appears it contains more state that just whether something can be targeted, but I've left it as is for now (and I'm to blame for the name anyway).


AsgardXIV:
Merged into #717

PR #718, opened 2023-12-04
Update RecipeNote

Haselnussbomber:
- Updated `RecipeNote.RecipeEntry` struct
  - Based on "E8 ?? ?? ?? ?? 8B 43 10 4D 8D 6D 24" and structs by ffxiv_exdgetters.py
- Added `RecipeNote.GetCraftTypeLevel(byte craftType)`
- Added static `QuestManager.IsRecipeComplete(uint recipeId)`

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
@@ -25,7 +25,7 @@ public unsafe partial struct PlayerState {
     [FieldOffset(0x7D)] public byte Tribe;
     [FieldOffset(0x7E)] public byte CurrentClassJobId;
 
-    [FieldOffset(0x80)] public void* CurrentClassJobRow;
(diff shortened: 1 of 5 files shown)


FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
```diff
@@ -25,6 +25,7 @@ public unsafe partial struct PlayerState {
     [FieldOffset(0x7D)] public byte Tribe;
     [FieldOffset(0x7E)] public byte CurrentClassJobId;
 
+    [FieldOffset(0x80)] public void* CurrentClassJobRow;
```
MidoriKami, line 28: Should this be nint? I believe we want to avoid void* where possible

Haselnussbomber: I don't mind either. I just thought I'd use void* because we don't have row structs.
Changed it to nint.

PR #721, opened 2023-12-09
Add AtkInputManager

Soreepeong:
Used to access clipboard the raw Utf8String with payloads intact, in case of copying from the game and then pasting; added the minimal to do that.

## Notes
* `+0x54cc50` deals with Ctrl+A/C/V/X; clipboard setting code from InputText can be found there.
* `+0xF70A70` deals with copying from ChatLog.


Haselnussbomber:
We have attributes and code generators that make it easier to add vfs, like this:
```cs
[VTableAddress("48 8D 05 ?? ?? ?? ?? 48 8B D9 48 89 01 48 83 C1 08 E8 ?? ?? ?? ?? 48 8D 4B 70 E8 ?? ?? ?? ?? 48 8B C3", 3)]
[StructLayout(LayoutKind.Explicit, Size = 0xD8)]
public unsafe partial struct ClipboardData
{
    [FieldOffset(0x8)] public Utf8String SystemClipboardText;
    [FieldOffset(0x70)] public Utf8String CopyStagingText;

    [VirtualFunction(1)]
    public partial void WriteToSystemClipboard(Utf8String* stringToCopy, Utf8String* copiedStringWithoutPayload);

    [VirtualFunction(2)]
    public partial Utf8String* GetSystemClipboardText();

    [VirtualFunction(3)]
    public partial void SetCopyStagingText(Utf8String* utf8String);

    [VirtualFunction(4)]
    public partial void ApplyCopyStagingText();
}
```

Also, have a look at `Client::System::Framework::Framework` + 0x2B68. It holds the same struct at 0x10, but is updated every time the clipboard data changes, because it's updated via the [WM_DRAWCLIPBOARD](https://learn.microsoft.com/en-us/windows/win32/dataxchg/wm-drawclipboard) case at 0x140057B62. 😉

Soreepeong:
Accessing one via AtkInputManager is desired atm, because it makes it possible to backtrack to the Utf8String that got copied, but with payloads. Otherwise, `ClipboardReadWrite` has been changed to `ClipboardData`, with the mentioned changes.

Soreepeong:
Will be done in an one go from #722.

PR #724, opened 2023-12-15
Fix function signatures added with #691

Haselnussbomber:
Somehow my comments on #691 were ignored by everyone. 😕
If this is wrong, then please explain it to me.


wolfcomp:
Ok if this was posted in discord it should also have been posted to the GitHub as it's impossible when it becomes over a week old to find said talk again.

Haselnussbomber:
I didn't talk about it on Discord. These were just comments in the PR here on GitHub. All good though!

PR #732, opened 2023-12-23
Add Character.GazeContainer

Haselnussbomber:
It looks like the struct at Character+0xCB0 is used to handle head/eye-tracking, so I called it `GazeContainer`, if that's alright.
I use `BannerHeadDirection` and `BannerEyesDirection` already and wanted to add them, but then I went exploring.
Sadly I couldn't make much sense of it, but at least I learned about the `/facecamera` command which has fields in there (local player only?!). 👀

---

Other small changes on `Character`:
- Copied the field `Character.Movement` (fixed bytes) to `Character.MovementBytes` and marked `Character.Movement` obsolete, so it can be used for a struct in the future. I can also just mark it obsolete and don't make a new field, if thats better.
- Copied `LookTargetId` into `GazeController.TargetInfo` and marked the old one obsolete.
- Changed FieldOffset of `StatusFlags4` to make it easier to see it's inside a struct that hasn't been added yet. Same offset.
- Removed/added new lines based on memory layout.

Also added some ctor functions and vtbls from other Containers to the data.yml.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Character/Character.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Character/Character.cs
@@ -30,14 +30,7 @@ public unsafe partial struct Character {
     [FieldOffset(0x970)] public ActionTimelineManager ActionTimelineManager;
     [FieldOffset(0xCB0)] public GazeContainer Gaze;
 
-    /// <summary>
-    /// The current target for this character's gaze. Can be set independently of soft or hard targets, and may be set
-    /// by NPCs or minions. For players, an action cast will generally target the LookTarget (which generally will be
-    /// the soft target if set, then the hard target).
-    /// </summary>
-    /// <remarks>
-    /// Unlike other GameObjectIDs, this one appears to be set to fully 0 if the player is not looking at anything.
-    /// </remarks>
+    /// <inheritdoc cref="GazeController.TargetInfo.LookTargetId"/>
     [Obsolete("Use Character.Gaze.Controller.Torso.TargetInfo.LookTargetId")]
     [FieldOffset(0xCB0 + 0x50)] public GameObjectID LookTargetId;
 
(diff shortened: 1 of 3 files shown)


Haselnussbomber:
CExporter wasn't happy, so I marked `LookTargetId` obsolete ("Use Character.Gaze.LookTargetId").

AsgardXIV:
I think this is more like

```
    [StructLayout(LayoutKind.Explicit, Size = 0x620)]
    public struct GazeContainer {
        [FieldOffset(0x00)] public void** ContainerVTable;
        [FieldOffset(0x08)] public Character* OwnerObject;
        [FieldOffset(0x10)] public GazeDriver Driver;
    }

    [StructLayout(LayoutKind.Explicit, Size = 0x5E0)]
    public struct GazeDriver {
        [FieldOffset(0x00)] public void** DriverVTable;
        [FieldOffset(0x10)] public Character* OwnerObject;
        [FieldOffset(0x20)] public Gaze Body;
        [FieldOffset(0x200)] public Gaze Head;
        [FieldOffset(0x3E0)] public Gaze Eyes;
        [FieldOffset(0x5C0)] public Gaze Extra;
    }

    [StructLayout(LayoutKind.Explicit, Size = 0x1E0)]
    public struct Gaze {
        [FieldOffset(0x00)] public void** GazeVTable;
        
    }
```

Haselnussbomber:
Yeah, make sense. I mapped it out in ReClass and saw it using an array of 3x 0x1E0 structs.
But if I add this I'm getting in trouble with CExporter again because of `LookTargetId`. 😕

screenshot

The last `Gaze Extra` makes no sense though, since its out of bounds (0x5C0 + 0x1E0 = 0x7A0 > 0x5E0).^^

AsgardXIV:
Oh, I just misread the constructor. You're right, it's only 3 entries. Corrected my comment.
I think @chirpxiv did some work in this area, and may know what's actually in the Gaze itself.

Haselnussbomber:
Alright, I made a separate GazeController with its inner structs. Should be fine now?

(Sorry for the multiple pushes, I got a bit confused for a second. 😅)

AsgardXIV:
Yep, looks good to me. Now at least the structure is captured, people can work out the fields as they need them.

PR #733, opened 2023-12-25
Implement STL containers with C# collection views

Soreepeong:
Continuation of #666, per @MidoriKami's request.

Tests are not done in a comprehensive way. It is possible that some parts of it has to be patched on use.

* Implements move/copy/default construction semantics for operations that modifies the struct or underlying buffer.
* Implements `IMemorySpace` specification using generic types.
* Implements `std::list`, `std::vector`, `std::basic_string`, `std::set`, `std::pair`, `std::map` for both read/write operations.
* Implements `std::deque` for read operations.
* `StdLinkedList` is not touched for the time being.

### Usage notes
If the contained type is not [trivial](https://en.cppreference.com/w/cpp/named_req/TrivialType), then any mutating operation will cause undefined behavior, until the contained type gets `IStaticNativeObjectOperation<T>` recursively implemented.


Soreepeong:
Decided that having a `PointerSpan<T>` would make things much more convenient.

aers:
We'll have to hold on merging this til an API bump presumably.

PR #734, opened 2023-12-25
Filling out RaptureHotbarModule

KazWolfe:
I swear one day we will know everything about this class. Using `internal` to mark things I'm not completely sure about yet.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/UI/Hotbar.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/UI/Hotbar.cs
@@ -1,3 +1,5 @@
+using FFXIVClientStructs.FFXIV.Client.UI.Misc;
+
 namespace FFXIVClientStructs.FFXIV.Client.Game.UI;
 
 [StructLayout(LayoutKind.Explicit, Size = 0x08)]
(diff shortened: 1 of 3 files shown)


KazWolfe:
The main thing that appears to be of interest here is `Client::Game::UI::Hotbar`. It implements `Client::UI::Misc::RaptureHotbarModule::ClearCallback`, which in turn appears to provide three vfuncs:

- vf1: `ResetAllHotbars`
- vf2: `ResetPvEHotbarsForClass`
- vf3: `ResetPvPHotbarsForClass`

These methods are responsible for initializing the "default" hotbar sets for each class. They appear to be called upon first run, certain game initialization steps, and other behavior. Calls to these three vfuncs are pretty entangled across `ReadFile` alongside a bunch of other variables that make things a bit complicated to track. In no particular order:

- `ModuleReady` seems to be set after `ReadFile` completes, after everything appears to have stabilized. When this is false, hotbar swaps do not happen.
- `DatFileLoadedSuccessfully` seems to be set after any successful read (but before any migrations?) are run as part of `ReadFile`. If reading the DAT file fails, this will be set to `false`.
- `ExpacJobHotbarsCreated` and `PvPHotbarsCreated` both appear to be migration-related jobs to fill out initial hotbars, but I'm not particularly clear as to when they're initialized or used.
- `ClearCallbackPresent` is likely misnamed, and seems to track whether to call the above-listed vfs from RaptureHotbarModule or not. Given that setting these fields is... complicated, again no idea when this actually happens in the lifecycle.

---

I also suspect that `UiHotbar` should really be `ClearCallback` of the same type, but that struct currently isn't defined and I'm not sure if that's the right way to handle this.

KazWolfe:
Something else: `RaptureHotbarModule` is getting big (no thanks to all my documentation comments)... It seems like it would make sense to move certain things into a child namespace, but we can't use `Client.UI.Misc.RaptureHotbarModule` as a namespace. Is there any preference for what to call this namespace, or should we just keep adding things to this one file?

aers:
Good stuff Kaz

PR #743, opened 2024-01-06
Rollback BGCollisionModule to use System.Numerics

KazWolfe:
Reverts an API break caused by a type change.

ref. https://github.com/goatcorp/Dalamud/pull/1600


Haselnussbomber:
I'm not sure if this has any implications in any way, but the `Common::Math::Vector3` struct had a size of 0x10, which aers changed it back to it after it was changed to 0xC in a PR.
See [history](https://github.com/aers/FFXIVClientStructs/commits/main/FFXIVClientStructs/FFXIV/Common/Math/Vector3.cs) of `FFXIV/Common/Math/Vector3.cs`.

goaaats:
If this is actually broken, we should fix it and DelvUI might just have to update, but it seems to be fine for them.

KazWolfe:
Yeah, I caught that, and un-reverted my revert to only catch the important things. Hopefully.

Haselnussbomber:
Yeah, CExporter check to fail is expected, because the fields are now overlapping in BGCollisionModule. Technically it shouldn't be a problem for now until we can add the BGCollisionModule version of Vector3 (with size 0xC).

aers:
Actually what happened here is they got changed to the BGCollision types but I knew that would be breaking so I changed them to what I thought they used to be (the CS types), I didn't realize they originally used the numerics types.

aers:
Note:

Common::Math doesn't actually exist. The numerics types in there belong to the Graphics namespace. I'm honestly not sure I remember why I agreed to have them moved.

The Graphics Vector3 is 16 bytes, with an extra 4 byte dummy field.

BGCollision has its own math classes, with a 12 byte Vector3.

aers:
Closing this since we fixed it elsewhere!

PR #744, opened 2024-01-07
Move RaptureHotbarModule things around

KazWolfe:
- Adds new `RaptureHotbarModule.ClearCallback` struct and vfuncs.
  - Move the vfuncs from `Client::Game::UI::Hotbar` to this struct, and mark it as inheriting.
  - Wrapped in a `partial struct RaptureHotbarModule` to match RTTI naming, as we can't create a `Client::UI::Misc::RaptureHotbarModule` namespace.
- Move the various RaptureHotbarModule sub-structs to their own files, based on concern.
  - These classes are all currently in the wrong namespace, which will need to be fixed in API 10 by wrapping them similar to how `ClearCallback` is handled.

For separation-by-concern, the following files were created in a (highly) opinionated manner:

- `RaptureHotbarModule.ClearCallback.cs` was created to hold `ClearCallback` (and only that).
- `RaptureHotbarModule.HotBar.cs` was created to hold `HotBar`.
  - `HotBarSlot` is not included there because that class is big.
- `RaptureHotbarModule.HotBarSlot.cs` was created to hold `HotBarSlot` and friends.
  - `DutyActionSlot` is included as it's too small to warrant its own file.
- `RaptureHotbarModule.HotBarSlotType.cs` was created to hold the enum with the same name. It was not included in `HotBarSlot` as `SavedHotBarSlot` also wants it.
- `RaptureHotbarModule.HotBarUiIntermediate.cs` was created to hold `HotBarUiIntermediate`.
  - I believe this class is part of RHM still, but lives between it and UI, hence the name. For now this is internal-only as I'm still not sure on a couple things, and it needs a doc pass.
- `RaptureHotbarModule.SavedHotBar.cs` was created to hold everything related to saved hotbars - `SavedHotBarGroup`, `SavedHotBar`, and `SavedHotBarSlot`. These classes are all independently too small to warrant their own files in my opinion.


FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureHotbarModule.ClearCallback.cs
```diff
@@ -0,0 +1,38 @@
+namespace FFXIVClientStructs.FFXIV.Client.UI.Misc;
+
+public partial struct RaptureHotbarModule {
+    [StructLayout(LayoutKind.Explicit, Size = 0x08)]
+    public unsafe partial struct ClearCallback {
+        [FieldOffset(0x0)] public void** vtbl;
```
wolfcomp, line 6: You should add a `, CExportIgnore()` for this, since it's exported from the VirtualFunction generator to the header files

FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureHotbarModule.HotBarSlot.cs
```diff
@@ -0,0 +1,307 @@
...
+
+// ToDo: Wrap in RaptureHotbarModule partial struct for namespacing (API 10)
+[StructLayout(LayoutKind.Explicit, Size = Size)]
+public unsafe partial struct HotBarSlot {
+    public const int Size = 0xE0;
+
+    /// The string that appears when a hotbar slot is hovered over.
+    ///
+    /// Calculated by concatenating GetDisplayNameForSlot with PopUpKeybindHint (in most cases).
+    [FieldOffset(0x00)] public Utf8String PopUpHelp; // TODO (apiX): Rename to DisplayName
```
Haselnussbomber, line 15: Why not make a new field and mark the old one obsolete?
```suggestion
    [FieldOffset(0x00), Obsolete("Use DisplayName")] public Utf8String PopUpHelp;
    /// <inheritdoc cref="PopUpHelp"/>
    [FieldOffset(0x00)] public Utf8String DisplayName;
```

KazWolfe: I'm still not certain if renaming the field is the right move. This field controls this specific value:

image

(Also, this is not a net-new field, that existed before this PR. I just moved the file.)

Haselnussbomber: Does this include the hotkey hint? Could probably be named `TooltipText` instead.

KazWolfe: Might be confused with the *actual* tooltip, which doesn't actually use that name at all:

image

Naming things hard, and I don't want to deal with it right now because it isn't net-new.

wolfcomp: Both are in theory a Tooltip, but I would probably keep the name as is since the other one is more of a full Tooltip than just the Action name and hotkey

KazWolfe: Removed pending rename due to maintainer request.

Haselnussbomber: One is AtkTooltip, the other is ItemDetail. 🤓☝️

aers:
Just to double check - no breaking changes right?

KazWolfe:
To the best of my knowledge, this should be API stable. I'll verify tonight however.

PR #746, opened 2024-01-08
Extend Framework

Koenari:
I would appreciate someone really looking this over (especially for typos and accidental api breakage)

Adds following classes
- ThreadManager
- Cursor
- GameWindow *
- File-Manager/-Thread-AccessPath

\* Acknowledgements:
- GameWindow is a completely assumed name and most likely is a much more low-level class

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/System/File/FileAccessPath.cs
+++ b/FFXIVClientStructs/FFXIV/Client/System/File/FileAccessPath.cs
@@ -1,0 +1,16 @@
+namespace FFXIVClientStructs.FFXIV.Client.System.File;
+
+[StructLayout(LayoutKind.Explicit, Size = 0x210)]
+public unsafe struct FileAccessPath {
+    [FieldOffset(0x000)] public fixed char Buffer[260];
+    [FieldOffset(0x208)] public char* LongStringPtr;
+
+    public string AsString {
+        get {
+            if (LongStringPtr == null)
+                fixed (char* p = Buffer)
+                    return new string(p);
+            else return new string(LongStringPtr);
+        }
+    }
+}
--- a/FFXIVClientStructs/FFXIV/Client/System/File/FileAccessPath.cs
+++ b/FFXIVClientStructs/FFXIV/Client/System/File/FileAccessPath.cs
@@ -5,12 +5,11 @@ public unsafe struct FileAccessPath {
     [FieldOffset(0x000)] public fixed char Buffer[260];
     [FieldOffset(0x208)] public char* LongStringPtr;
 
-    public string AsString {
-        get {
-            if (LongStringPtr == null)
-                fixed (char* p = Buffer)
-                    return new string(p);
-            else return new string(LongStringPtr);
-        }
+    public override string ToString() {
+        if (LongStringPtr == null)
+            fixed (char* p = Buffer)
+                return new string(p);
+        else
+            return new string(LongStringPtr);
     }
 }
--- a/FFXIVClientStructs/FFXIV/Client/System/Framework/Framework.cs
+++ b/FFXIVClientStructs/FFXIV/Client/System/Framework/Framework.cs
@@ -1,6 +1,7 @@
 using FFXIVClientStructs.FFXIV.Client.Game;
 using FFXIVClientStructs.FFXIV.Client.Network;
 using FFXIVClientStructs.FFXIV.Client.System.Configuration;
+using FFXIVClientStructs.FFXIV.Client.System.File;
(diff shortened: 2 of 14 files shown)


FFXIVClientStructs/FFXIV/Client/System/File/FileAccessPath.cs
```diff
@@ -0,0 +1,16 @@
...
+    [FieldOffset(0x208)] public char* LongStringPtr;
+
+    public string AsString {
+        get {
+            if (LongStringPtr == null)
+                fixed (char* p = Buffer)
+                    return new string(p);
+            else return new string(LongStringPtr);
+        }
+    }
```
Haselnussbomber, line 15: I'd suggest overriding C#'s `ToString()` instead.
```suggestion
    public override string ToString() {
        if (LongStringPtr == null)
            fixed (char* p = Buffer)
                return new string(p);
        else
            return new string(LongStringPtr);
    }
```

Haselnussbomber (commented):
Couple suggestions from me. 🙂

ida/data.yml
```diff
@@ -2233,6 +2261,17 @@ classes:
...
+    vtbls:
+      - ea: 0x1419B0378
+    vfuns:
+      0: dtor
+      4: SetCursorTyp
+      5: SetIsCursorVisible
+      8: SetUSeSoftwareCursor
+      9: GetUSeSoftwareCursor
+      14: IsCursorVisible
+      21: GetIsMouseNotCaptured
```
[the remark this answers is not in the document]
Haselnussbomber: Edited this. Autopilot brain renamed them to Enable/DisableSoftwareCursor, but they aren't doing that, sorry. 🤦‍♂️
Fixed the typo though. 🙂

FFXIVClientStructs/FFXIV/Client/System/Framework/Framework.cs
```diff
@@ -39,7 +51,9 @@ public unsafe partial struct Framework {
     [FieldOffset(0x17C4)] public float FrameRate;
     [FieldOffset(0x17D0)] public bool WindowInactive;
 
-    [FieldOffset(0x220C)] private fixed char userPath[260]; // WideChar Array
+    [FieldOffset(0x19EC)] private fixed char gamePath[260];
+    [FieldOffset(0x1DFC)] private fixed char sqPackPath[260];
+    [FieldOffset(0x220C)] private fixed char savePath[260];
```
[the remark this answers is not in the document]
Haselnussbomber: Oh, it seems like these are UTF16 and the FixedStringGenerator only supports UTF8, so this won't work.

ida/data.yml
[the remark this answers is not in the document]
Koenari: Yeah those two are jsut Get/Set for the field

FFXIVClientStructs/FFXIV/Client/System/Framework/Framework.cs
```diff
@@ -65,11 +81,26 @@ public unsafe partial struct Framework {
     [MemberFunction("E8 ?? ?? ?? ?? 89 47 2C")]
     public static partial long GetServerTime();
 
-    public string UserPath {
-        get {
-            fixed (char* p = userPath) {
-                return new string(p);
-            }
+    [Obsolete("Use SavePath")]public string UserPath => SavePath;
```
Caraxi, line 84: Why rename it? The launch parameter it is set by is called UserPath

Koenari: I did miss that and opted for a more descriptive name. Is reverted now as you are correct and we should adhere to SE's naming

FFXIVClientStructs/FFXIV/Client/System/Framework/Framework.cs
```diff
@@ -50,6 +64,8 @@ public unsafe partial struct Framework {
 
     [FieldOffset(0x2BF0)] public GameVersion GameVersion;
 
+    [FieldOffset(0x35C0)] public void* SteamApiLibrary;
+
```
Haselnussbomber, line 68: `SteamApiLibrary` can be omitted, PR #748 is going to add this. 🙂
```suggestion
```

Koenari: Not hte same field, this is the DLL 0x35B8 is the wrapper

Haselnussbomber: I see, there is a mistake in the other PR. Sorry.

PR #755, opened 2024-01-12
Update InventoryType and InventoryItem

Haselnussbomber:
- Updates the `InventoryType` enum
  - Taken from an internal array at 0x141A1E0B0.
  - Type 2006 and 2008 are still unknown. I assume one of them is for trading with other players, and the other for getting items repaired by other players.
  - There is also an id 9999, which i haven't added. It's a special id to indicate an invalid or unset container id. Not sure on the naming of that one.
  - In addition, there are ids 2000 and 2013 that I have not added. I think these are start and end markers for the types in the 2000 range.
- Updates `InventoryItem` to support links.
  - Allows InventoryItem to reference a specific, other InventoryItem based on the container type and slot id.
    - In the case of a symbolic InventoryItem, the ItemId field is unused. Instead, it is divided into two ushorts: one for the container (InventoryType) and another for the slot id.
  - Added multiple native getter functions for the various fields, which all support resolving the linked InventoryItem.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
@@ -123,11 +123,17 @@ public enum InventoryType : uint {
 
     Currency = 2000,
     Crystals = 2001,
+    MailEdit = 2002, // used by LetterEditor
     Mail = 2003,
     KeyItems = 2004,
     HandIn = 2005,
+    Unknown2006 = 2006,
     DamagedGear = 2007,
+    Unknown2008 = 2008,
     Examine = 2009,
+    Reclaim = 2010, // LegacyItemStorage, HousingWithdrawStorage
+    HousingExteriorAppearanceEdit = 2011,
+    HousingInteriorAppearanceEdit = 2012,
     ReconstructionBuyback = 2013, // Doman Enclave Reconstruction Reclamation Box
 
     ArmoryOffHand = 3200,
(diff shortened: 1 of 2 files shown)


FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
```diff
@@ -123,11 +195,17 @@ public enum InventoryType : uint {
 
     Currency = 2000,
     Crystals = 2001,
+    MailEdit = 2002, // used by LetterEditor
     Mail = 2003,
     KeyItems = 2004,
     HandIn = 2005,
+    Unknown2006 = 2006,
     DamagedGear = 2007,
+    Unknown2008 = 2008,
```
Koenari, line 204:
```suggestion
    DamagedGear = 2007,
```
I don't think you meant to commit those

Haselnussbomber: Yes, I did intend to commit them, becuase they are valid ids in the games array.
Like I said, they are probably for trading and reparing with other players.
Besides them and the 9999, there is also a 2000 and 2013 probably used for start and end marking of these types, commonly found in enums.
Whenever the correct purpose is found, they can be marked obsolete and point to the correct name. 🙂

Koenari: Yeah, sorry I should go to bed and review when awake actually

PR #767, opened 2024-01-13
Revert API breakage in ConfigBase

KazWolfe:
Proposed fix for a noted API break on a possibly-critical method.


aers:
I don't want to revert this unless you can point out someone actually using it.

KazWolfe:
To my knowledge, no mainline plugins use this API, and neither does Dalamud. I cannot make any guarantees about custom plugins at this time.

I'll pull this PR for now.

PR #768, opened 2024-01-13
Expose pointers to Steam interfaces

KazWolfe:
This is more of a proposal than a true PR, seeing as this is rather non-CS-y.

The current implementation of the Steam API documentation does not provide any way to receive the game's assigned instances of `ISteamUtils`, `ISteamApps`, etc. This means that anyone wishing to consume these will need to make their own instances, which may or may not end well. This pull requests adds a proposed method to fetch these pre-existing interface pointers for use elsewhere.

n.b. Steam API encourages using `SteamInternal_ContextInit` to fetch an interface reference from the provided context. CS currently does not provide a way to grab this function pointer so a stubbed method was used to just return the cached reference. This *should* be safe, but is less than optimal.


aers:
I'll take a look at this in a few days

KazWolfe:
Marking as draft as I found a couple other interesting-looking structs I want to document first.

aers:
just an fyi im a bit sick right now so i probably won't be looking at this super soon, but you could get pohky or someone esle to

FFXIVClientStructs/FFXIV/Client/System/Input/SoftKeyboards/SteamGamepadTextInput.cs
```diff
@@ -0,0 +1,9 @@
+namespace FFXIVClientStructs.FFXIV.Client.System.Input.SoftKeyboards;
+
+[StructLayout(LayoutKind.Explicit, Size = 0x18)]
+public unsafe partial struct SteamGamepadTextInput {
+    [FieldOffset(0x00)] public SoftKeyboardDevice* PlatformInputMethod;
+
+    [FieldOffset(0x08)] public byte Enabled;
+    [FieldOffset(0x10)] public void* TargetSoftKeyboardInputInterface; // populated as part of 140577A90
```
hedgeye-fslezak, line 8: change to a `SoftKeyboardDeviceInterface.SoftKeyboardInputInterface*`

FFXIVClientStructs/FFXIV/Client/System/Input/SoftKeyboardDevice.cs
```diff
@@ -0,0 +1,30 @@
...
+
+[StructLayout(LayoutKind.Explicit, Size = 0x08)]
+public unsafe partial struct SoftKeyboardDevice {
+    [FieldOffset(0x00), CExportIgnore] public void** vtbl;
+
+    [VirtualFunction(0)] 
+    public partial void Dtor(bool freeMemory);
+    
+    [VirtualFunction(1)]
+    public partial bool Initialize();
```
hedgeye-fslezak, line 11: Rename to `Enable`

KazWolfe, line 11: Rename to `Enable`

FFXIVClientStructs/FFXIV/Client/System/Input/SoftKeyboards/SteamGamepadTextInput.cs
```diff
@@ -0,0 +1,9 @@
...
+}
```
KazWolfe, line 9: Rename this struct to `SteamGamepadSoftKeyboard`, rename `TargetSoftKeyboardInputInterface` to `TargetInputInterface` and set concrete type.

FFXIVClientStructs/FFXIV/Component/SteamApi/SteamTypes.cs
```diff
@@ -23,4 +27,30 @@ public struct CSteamId {
...
+        /// <summary>
+        /// Gets a pointer to the Steam API instance stored in this context.
+        /// This pointer is suitable for use in imported Steam API methods.
+        /// </summary>
+        public nint GetInterface() {
+            fixed (SteamInterfaceContext* ptr = &this) {
+                // ToDo: Actually call `SteamInternal_ContextInit(ctx)` here, rather than just returning the cached pointer.
+                //       Needs some way to get the fptr for that method first. For now this is *probably fine* (and won't blow up)
+                //       but isn't awesome.
+                return (nint)ptr->CachedInterfacePtr;
```
KazWolfe, line 52: This would require some way for ClientStructs to be able to call a method from the import table by reference. I don't think this is particularly useful to have except here, but it may be a good thing to have in general.

PR #772, opened 2024-01-13
Add RaptureTextModule.FormatName and helpers

Haselnussbomber:
I figured most people struggle with getting properly formatted names, especially for languages other than english.
This global, static function seems to be a helper to get the correct parameters for FormatAddonText to format a variety of names (Items, BNpcNames, ActionNames etc.).
Unfortunately, there is no support for plural forms, since adjusting `intParam2` (passed to FormatAddonText) doesn't seem to change anything.
I added helper functions with combinations that worked in my testing.

Like I said, it's a global function, and locaded very early on in the exe. It's probably not part of RaptureTextModule, but I didn't know where else to put it and since it uses RaptureTextModule, I thought it might fit there.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureTextModule.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureTextModule.cs
@@ -44,6 +44,8 @@ public unsafe partial struct RaptureTextModule {
 
     public enum NameFormatterIdConverter : uint
     {
+        None = 0,
+
         // ObjStr
         ObjStr_BNpcName = 2,
         ObjStr_ENpcResident = 3,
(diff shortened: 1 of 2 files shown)


aers:
It's a generic static UI function like PlaySoundEffect, GetUIColor, etc. (same location in the binary).

PlaySound is in UIModule, I guess, but I don't entirely object to this one being here.

Are you sure the params are mixed int/uint?

Haselnussbomber:
I'm not sure. Its probably both `int` since its calling `FormatAddonText` (`FormatAddonText2` in C#).

aers:
Yeah I think all args as int is probably more correct.

Haselnussbomber:
I don't mind.

---

While we're at the subject of statics, I also found this one:
```cs
[MemberFunction("80 F9 07 77 10")]
public static partial bool IsGatheringPointRare(byte gatheringPointType);
```
Maybe we should have a `Statics` struct? Idk.

aers:
Its a consideration. All the static functions in the same region as PlaySoundEffect and this name function (1400Axxxx iirc) are UI helper code and belong to some global UI class.

PR #780, opened 2024-01-20
Bit of project cleanup 🧹 + Buddy changes and addition of TitleList

Haselnussbomber:
I took the time to clean up/change a few things:

- changed the comments to have roughly the same format
  - removed `size = 0x...` comments, which I find redundant to the StructLayout Size param
  - put all signatures in apostrophs, so they can be scanned with annas sig_checker.py
  - changed spacing around the comments, so they are sticking to the struct instead of the namespace
- moved enums to the bottom of a struct/file
- moved `Instance()` up, before the fields
- removed unused `usings` (automated)
- removed some semicolons after structs
- moved `ActionTimelineDriver` into its own file
- in data.yml:
	- grouped UIState structs together (in order) and converted some of them from `globals` to `classes` entries
	- added some functions that were only present in C#
	- renamed some functions in data.yml that had a different name in C#

The only thing I actually changed was `IsBuddyEquipUnlocked`. It's calls `private IsBuddyEquipUnlockedInternal`, but passed the wrong parameter, because it adds +0xC to a1, which is the space between `CompanionPtr` and `BuddyEquipUnlockBitmask`. 🤔 Would be nice if someone could double check that implementation. 🤷‍♂️


FFXIVClientStructs/FFXIV/Client/Game/InstanceContent/InstanceContentDeepDungeon.cs
```diff
@@ -1,5 +1,12 @@
 namespace FFXIVClientStructs.FFXIV.Client.Game.InstanceContent;
 
+// Client::Game::InstanceContent::InstanceContentDeepDungeon
+//   Client::Game::InstanceContent::InstanceContentDirector
+//     Client::Game::InstanceContent::ContentDirector
+//       Client::Game::Event::Director
+//         Client::Game::Event::LuaEventHandler
+//           Client::Game::Event::EventHandler
+// ctor "E9 ?? ?? ?? ?? 3D ?? ?? ?? ?? 77 73 3D"
```
Nukoooo, line 9: I would suggest using direct signature for InstanceContent ctor, because it is likely to be invalid/wrong if SE adds new stuff to the function where they construct or the compiler optimization
The signature I made from 6.50 binary
``48 89 5C 24 ?? 48 89 6C 24 ?? 48 89 74 24 ?? 57 48 83 EC ?? 48 8B F9 E8 ?? ?? ?? ?? C6 87``

FFXIVClientStructs/FFXIV/Client/Game/UI/Buddy.cs
```diff
@@ -45,11 +46,10 @@ public struct BuddyMember {
...
-        fixed (byte* p = BuddyEquipUnlockBitmask) {
-            return IsBuddyEquipUnlockedInternal(p, buddyEquipId);
-        }
+        fixed (BuddyMember** ptr = &CompanionPtr)
+            return IsBuddyEquipUnlockedInternal(ptr, buddyEquipId);
     }
 
     [MemberFunction("E9 ?? ?? ?? ?? 0F B7 50 02 41 B8")]
-    private static partial bool IsBuddyEquipUnlockedInternal(void* ptr, uint buddyEquipId);
+    private static partial bool IsBuddyEquipUnlockedInternal(BuddyMember** companionPtr, uint buddyEquipId);
```
wolfcomp, line 54: Unsure if the generator spits this out as a public member still, but if so this should not be used in terms of plugins.

wolfcomp (approved):

FFXIVClientStructs/FFXIV/Client/Game/UI/Buddy.cs
Haselnussbomber: what do you mean by "this should not be used in terms of plugins"? it's a private function, only accessable to `IsBuddyEquipUnlocked`. plugins can't use it

Haselnussbomber: Oh, I see, you mean this, right?

FBHEpJ9SDH

Only the address is public, of course.

Haselnussbomber:
I'm getting scared of merge conflicts. Hopefully someone merges this soon. 😅

PR #787, opened 2024-01-24
Add Utf8String.ToUpper

Haselnussbomber:
Couldn't find out what a4 does. 🤷‍♂️
Also, `everyWord` is only handled when `firstCharOnly` is true, otherwise the whole text is uppercase anyway.


aers:
Sure looks like its converting characters to capital vowels... for what purpose

PR #788, opened 2024-01-25
Obsolete FixedArrayAttribute

Haselnussbomber:
`FixedArrayAttribute` doesn't seem to be used anywhere and I fell into the trap of using it instead of `FixedSizeArrayAttribute`.


aers:
I..honestly can't remember why this is here.

PR #789, opened 2024-01-25
Add PlayerState.ActiveFestivalIds/ActiveFestivalPhases

Haselnussbomber:
The lua function `IsOpenedFestival` (1408ED2E0) reads active festival ids from PlayerState. I assume the phases come right after, but can't tell for sure just yet. ~~Moogle Treasure Trove: The Second Hunt for Genisis should theoretically use the second phase of the same event. I'll leave this draft here until I can confirm it.~~

That said, `GameMain.ActiveFestivals` is not a `uint` array. Each festival entry has a ushort Id and a ushort Phase. Multiple xrefs confirm this, including the lua function `GetFestivalPhase` (1407D9ED0).   Sadly I can't add the FixedSizeArrayAttribute, because that only works on fixed byte arrays. I assume `GameMain.QueuedFestivals` has the same structure.


Haselnussbomber:
As it turns out, the Moogle Treasure Trove is not a festival at all. 🤷‍♂️

PR #793, opened 2024-01-28
Add AtkComponentTextInput.SetText

Haselnussbomber:


Haselnussbomber:
The address seemed to be part of AtkComponentInputBase, but it accesses offsets of a1 outside of it's size, so it has to be AtkComponentTextInput. 🤔

PR #804, opened 2024-02-15
Add TextParameter

Haselnussbomber:
I've been reading PlayerParameters (gnum/gstr) from GlobalParameters with success.
Now all we need is a way to make StdDeques ourselves, so we can pass local parameters to FormatString. 😅

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/Text/TextModule.cs
+++ b/FFXIVClientStructs/FFXIV/Component/Text/TextModule.cs
@@ -23,6 +23,6 @@ public unsafe partial struct TextModule {
     [VirtualFunction(9)]
     public partial Utf8String* ProcessMacroCode(Utf8String* output, byte* input);
 
-    [VirtualFunction(16)] //args is a pointer to a deque with the args, if any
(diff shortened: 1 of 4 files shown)


FFXIVClientStructs/FFXIV/Component/Text/TextModule.cs
```diff
@@ -23,6 +23,10 @@ public unsafe partial struct TextModule {
     [VirtualFunction(9)]
     public partial Utf8String* ProcessMacroCode(Utf8String* output, byte* input);
 
-    [VirtualFunction(16)] //args is a pointer to a deque with the args, if any
-    public partial bool FormatString(byte* input, void* args, Utf8String* output);
+    [Obsolete("Use FormatString with StdDeque<TextParameter>*")]
+    public bool FormatString(byte* input, void* localParameters, Utf8String* output)
```
pohky, line 27: you can remove this overload instead of obsoleting it, that's fine here

Haselnussbomber: If you say so. Okayge

PR #806, opened 2024-02-18
Fix tm days being off by 1

Haselnussbomber:
Seems like I f*d it up. Is this correct now?

**tm_mday**: day of the month – [1, 31]
**[DateTime.Day](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.day?view=net-8.0)**: The day component, expressed as a value between 1 and 31.

**tm_wday**: days since Sunday – [​0​, 6]
**[DateTime.DayOfWeek](https://learn.microsoft.com/en-us/dotnet/api/system.dayofweek?view=net-8.0)**: Sunday 0 - Saturday 6

https://en.cppreference.com/w/cpp/chrono/c/tm


Haselnussbomber:
I take that as a yes. 🤔

PR #812, opened 2024-02-24
Add a null check to `UldManager.SearchNodeById`

Caraxi:
Calling SearchNodeById before the UldManager has finished loading crashes the game.
This aims to avoid that

PR because idk


Haselnussbomber:
Why not just check `LoadedState == AtkLoadState.Loaded` before calling it? 🤔

Caraxi:
could do that too, but this guarantees the object list isn't going to terminate the game with a null reference

wolfcomp:
Weird that this crashes the game with what is in the decompiled view
img

Caraxi:
Hmm you're right, it does null check for itself

doesn't really make sense that its crashing at 
```
  [0]	ffxiv_dx11.exe+5376FC	(Component::GUI::AtkUldManager.SeardNodeById+0x1C)
```
which is
```
NodeCount = Objects->NodeCount;

PR #813, opened 2024-02-25
Add safety check for `AtkUldManager.SearchNodeById`

wolfcomp:


Haselnussbomber:
I don’t think CS should have such checks. Plugins should check the state before calling functions, otherwise we end up adding checks for every single function in CS.

Koenari:
To be again the pedantic one. This needs to be considered a breaking as the behavior and therefore return values of the function change in a way that could have an unknown impact on consumers. Even if I would consider the possibility of breaking any plugin very low. The correct way would be leaving old functionality and introducing a "SearchNodeByIdSafeguarded" and a note in the documentation comment on the original function.
I think this is a good opportunity to find and write down guidelines on how to properly indicate functions that are not part of the original executable and also a project wide stance on what should be part of CS.
While I understand that this has little short-term value it could be a significant improvement for the long term health of the project especially regarding new contributors.

PR #814, opened 2024-02-25
Update GHA to automate a bit more of checks

wolfcomp:
Updated CExporter action to comment with what is wrong and add the requested changes label.
Added ABI check between target and source branches of pr and automatically assign tag if breakage happens.

### Notes
These currently won't work due to some weirdness about how permissions are handled, and it's only getting read instead of write access.
Both actions only add but never remove the tags that have been added if a workflow finds something that needs to be changed or there is an ABI break, but then it's fixed later.

### DIFF
--- a/.github/workflows/pr-breaking-change-check.yaml
+++ b/.github/workflows/pr-breaking-change-check.yaml
@@ -1,7 +1,11 @@
 name: Breaking Change Check
 
 on:
-  pull_request:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
     branches:
       - main
     paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-breaking-change-check.yaml
```diff
@@ -0,0 +1,49 @@
...
+      run: git checkout ${{ github.head_ref }}
+    - name: Install dependencies for PR branch
+      run: dotnet restore
+    - name: Build PR branch
+      run: dotnet build FFXIVClientStructs/FFXIVClientStructs.csproj --output ida/cs-pr/
+    - name: Check for breaking changes
+      id: breaking_changes
+      run: |
+        dotnet tool install -g Microsoft.DotNet.ApiCompat.Tool
+        testout=$(apicompat -l ida/cs-main/FFXIVClientStructs.dll -r ida/cs-pr/FFXIVClientStructs.dll)
```
KazWolfe, line 31: Problems are pushed to STDERR, success message (if any) is pushed to STDOUT. This is specifically why we are testing the `"APICompat ran successfully without finding any breaking changes."` string, at least on a Windows runner.

PR #816, opened 2024-02-26
Some ContextMenu sigs

WorkingRobot:
Gonna use these for https://github.com/goatcorp/Dalamud/pull/1682


FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
```diff
@@ -42,6 +42,12 @@ public unsafe partial struct RaptureAtkModule {
     [MemberFunction("E8 ?? ?? ?? ?? 48 8B 46 58 48 85 C0")]
     public partial bool DecRefStringArrayData(int index);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 66 89 46 50")]
+    public partial nint OpenAddon(uint addonNameId, uint valueCount, AtkValue* values, AgentInterface* parentAgent, long unk, uint ownerAddonId, int unk2);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 0F B7 C0 48 83 C4 60")]
+    public partial nint OpenAddonByAgent(byte* addonName, AtkUnitBase* addon, int valueCount, AtkValue* values, AgentInterface* agent, nint a7, ushort parentAddonId);
```
wolfcomp, line 49: Please add this to `data.yml`

wolfcomp, line 49: Return type is `void` not `nint`

wolfcomp (commented):
There is probably more that is wrong, but this is the first pass I have done.

FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
Haselnussbomber: As far as I know, they addon id is always `ushort`.

Its also being passed like this (as word):
Q6wZRCtVu4
Which means the `AddonId` field in the [AgentInterface](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInterface.cs#L12) struct has the wrong type.

FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
```diff
@@ -43,10 +43,10 @@ public unsafe partial struct RaptureAtkModule {
     public partial bool DecRefStringArrayData(int index);
 
     [MemberFunction("E8 ?? ?? ?? ?? 66 89 46 50")]
-    public partial nint OpenAddon(uint addonNameId, uint valueCount, AtkValue* values, AgentInterface* parentAgent, long unk, uint ownerAddonId, int unk2);
+    public partial ushort OpenAddon(ushort addonNameId, uint valueCount, AtkValue* values, AgentInterface* parentAgent, ulong unk, ushort parentAddonId, int unk2);
```
Haselnussbomber, line 46: No, `addonNameId` should stay `uint`. There are currently 852 addon names it can pick from.

Haselnussbomber (changes requested):

FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
WorkingRobot: 852 is still below 65k, no?

Haselnussbomber: Oh right, lol. It's used as uint, but whatever I guess?

Haselnussbomber: I am not sure what happens on a CPU. Like, does it only write 2 bytes into the register if we specify it as ushort? That could be problematic if it reads an uint afterwards?!

WorkingRobot: Technically, the assembly code reads the register as a 32 bit value and moves it to a 64 bit value. So, I think it's a better idea to make it a uint just in case? But, when we're calling it from c#, I'm quite sure that the clr zeroes the whole 64-bit register first. I'll change it back to a uint..

FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
wolfcomp: Yeah, as I said in the main comment it was the first pass and I did it in maybe 5 to 10 minutes, so it was very rough.

PR #817, opened 2024-02-26
ModelRenderer stuff

Exter-N:
Not sure about whether having ShaderSceneKey and ShaderSubViewKey be distinct was the right thing to do: they have different vtables, but are structurally identical to the best of my knowledge (same number of vfuncs, same memory layout, same member semantics).

Also, not sure about `…::Kernel::Shader` actually being a superclass in its own right, distinct from what I had called `PVShader`.


FFXIVClientStructs/FFXIV/Client/Graphics/Kernel/ShaderSubViewKey.cs
wolfcomp, line 1: I would just use the other struct if these are 100% the same in a memory map standpoint since that indicates they use the same struct most likely in their own code

aers: > I would just use the other struct if these are 100% the same in a memory map standpoint since that indicates they use the same struct most likely in their own code

What? If they have different vtbls they are different classes.

PR #819, opened 2024-03-03
Assorted Data.yml changes

Koenari:
A lot of smaller and bigger stuff I found along my other work
Separated into a single PR for easier review and merge

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -6,11 +6,17 @@ globals:
   0x1419BE410: g_UIColorTable
   0x141A06EC8: g_HUDScaleTable
   0x141FDC380: g_ClientInstanceLimit
+  0x141FDC410: g_WindowStyle
+  0x141FDD498: g_FPS
   0x141FE0D38: g_LanguageCharArr # byte[] no pointer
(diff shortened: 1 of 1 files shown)


ida/data.yml
```diff
@@ -680,16 +731,22 @@ classes:
       0x1412E2FC0: CreateEventObject
       0x1412E30A0: CreateAreaObject
       0x1412E3190: DeleteObjectByIndex
+      x1412E32F0: DeleteAllObjects
```
Haselnussbomber, line 734:
```suggestion
      0x1412E32F0: DeleteAllObjects
```

Haselnussbomber (changes requested):

wolfcomp (changes requested):
Most things seem good, just some tiny things that need to pick about, mainly to try and get consistency and how c++ actually gets compiled where vfunc 0 is the `dtor` and the sub function call for consistency’s sake should be named `Finalize`

If you are confused why this is the case look at `Framework` or any of the `InfoProxy` definitions

pohky (changes requested):

PR #826, opened 2024-03-18
Add GoldSaucer manager and directors

Haselnussbomber:
Not sure about namespaces, I took inspiration from existing ones.

Should we even add unused ctors that got inlined in the data.yml? I imagine that the update script will do weird stuff with those.

Anyway, I haven't spent too much time on it (which is why there are a couple internal fields), so... more discoveries as they come, I guess. 🙂


Haselnussbomber:
Forgot to update the struct size.

Haselnussbomber:
I just realized I didn't check `GoldSaucerManager_GetInstance` xrefs (a bunch of lua functions use that, so we get some field names!)...
I'll do that tomorrow, it's past 5 am. ._.

PR #827, opened 2024-03-18
Pass Vector3 pointers to Raycast functions

Haselnussbomber:
This is a confusing one, because it works, even though it shouldn't.
According to Pohky, C# (the caller) creates a copy of `Vector3` on the stack, because `struct` is a value type and too big to be passed via the register itself. It passes it as a "stack pointer" instead, which happens to be exactly what the game needs - a pointer.
For me, it simply should be a `Vector3*`, because this is what I see in the games assembly - a dereference.
That it gets converted automatically and becomes a pointer is just ridiculous in my opinion, and should not be relied on, and instead the types should be fixed.

asm

Discussion in #plugin-dev where I try to understand what is happening (not very useful and basically explained above): https://discord.com/channels/581875019861328007/653504487352303619/1219332724792229900

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Scene/Camera.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Scene/Camera.cs
@@ -1,6 +1,7 @@
 using FFXIVClientStructs.FFXIV.Client.Graphics.Kernel;
 using FFXIVClientStructs.FFXIV.Common.Component.BGCollision;
 using FFXIVClientStructs.FFXIV.Common.Math;
+using SystemVec3 = System.Numerics.Vector3;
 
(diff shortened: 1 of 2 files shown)


Haselnussbomber:
Hello? Where is the other commit??
https://github.com/Haselnussbomber/FFXIVClientStructs/commit/70ac3b8c173f3d7120e0da87cf35f0f18427cfb8

*Edit:* Oh..
NsDgmVVamA

aers:
In the original game code, the parameters are likely passed by-reference (so the function signature is using &), so they should probably be pointers here..? I think you're actually right..

Haselnussbomber:
Leaving this open then for Dawntrail.

PR #830, opened 2024-03-23
Add FormatAddonText overloads

Haselnussbomber:
The third parameter of `FormatAddonTextApply` is the reason for the function names `FormatAddonText1`/`FormatAddonText2`.
If anyone finds out what the parameter does, we can maybe properly name those functions.


Haselnussbomber:
Whoops, used the wrong term. Those are overloads, obviously.

PR #835, opened 2024-03-28
Cexporter full rework

wolfcomp:
# Changes:
* Converts all structs to one unified file
* Implements a script for IDA to import this new struct file
* Implements a script for Ghidra to import this new struct file <sub>Script file is not feature complete</sub>


wolfcomp:
CExporter says it fails due to it not using the workflow that is in the pr but what is on main.

wolfcomp:
holy it actually took the updated script....

PR #848, opened 2024-04-13
Add AgentMap.EventMarkers

Haselnussbomber:
For @MidoriKami 🙂

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Event/Director.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Event/Director.cs
@@ -1,3 +1,4 @@
+using FFXIVClientStructs.FFXIV.Client.Game.UI;
 using FFXIVClientStructs.FFXIV.Client.System.String;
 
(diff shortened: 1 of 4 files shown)


MidoriKami:
Did you determine whether calling that function directly triggered network requests?

Haselnussbomber:
The `Client::Game::Event::Director.PopulateMapMarkers` functions don't do network requests.

---

In `sub_1408731E0` it iterates `Client::Game::Event::EventFramework.DirectorModule.DirectorList` and calls `PopulateMapMarkers` for every director to compile a list of all markers.

For the map it is called via `AgentMap_Update` (-> `sub_140BFCAF0` -> `sub_140BF2750` -> `sub_1408731E0`), but only when the map is open.
For the minimap it is called via `AgentHUD_Update` (-> `sub_140C0D4B0` -> `sub_140BF2750` -> `sub_1408731E0`) every frame.

MidoriKami:
:surprisedpikachu: I can't wait to play around with this thanks!

PR #849, opened 2024-04-13
Yet more bone simulator fields

redstrate:
A few more fields I found that are somewhat useful, I guess

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
@@ -8,10 +8,28 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Physics;
 [StructLayout(LayoutKind.Explicit, Size = 0x100)]
 public unsafe struct BoneSimulator {
     [FieldOffset(0x00)] public void* vtbl;
-    [FieldOffset(0x10)] public uint PhysicsGroup;
+
+    /// <summary> Non-exhaustive list of physics groups </summary>
+    public enum PhysicsGroup : uint
+    {
+        Clothing = 2,
+        HairA = 3, // Usually the hair's bangs
+        HairB = 4, // Typically the back of the hair
+        HairC = 5, // The sides of the hair
+        HairD = 6, // Extraneous hair bits
+        Chest = 7,
+        Earrings = 8,
+        Ears = 18,
+    }
+
+    [FieldOffset(0x10)] public PhysicsGroup Group;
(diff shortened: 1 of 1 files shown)


FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
```diff
@@ -8,7 +8,21 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Physics;
...
+        HairA = 3, // Usually the hair's bangs
+        HairB = 4, // Typically the back of the hair
+        HairC = 5, // The sides of the hair
+        HairD = 6, // Extraneous hair bits
+        Chest = 7,
+        Earrings = 8,
+        Ears = 18,
+    }
+
+    [FieldOffset(0x10)] public PhysicsGroup Group;
```
redstrate, line 25: Don't know the rules against API changes, is it OK to break? (I'm not sure if there's even any consumers)

FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BonePhysicsModule.cs
```diff
@@ -18,6 +18,7 @@ public struct BoneSimulators {
 public unsafe struct BonePhysicsModule {
     [FieldOffset(0x00)] public void* vtbl;
     [FieldOffset(0x10)] public Matrix4x4 SkeletonWorldMatrix;
+    [FieldOffset(0x40)] public Vector3 CharacterPosition;
```
wolfcomp, line 21: This is part of the `Matrix4x4` if you want to have just the vector3 component out, make a getter function for this instead.

wolfcomp (changes requested):

FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
wolfcomp: ABI changes will be auto marked as breaking change due to how C# works, even if there are no consumers for this. This is due to the check system is just comparing it between the old and new CS versions.

FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BonePhysicsModule.cs
redstrate: Oops yes, I miscalculated some of the sizes. I reverted this bit

FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
redstrate: Ah ok, cool!

PR #850, opened 2024-04-13
Add Agent getters to data.yml

Haselnussbomber:
This PR adds agent getter functions to the data.yml.

The following ones do not have any callers, so I left them out (unless I should add them? seems pointless 🤔 ):

```
0x140225660: GetAgentScenarioTree
0x140225670: GetAgentMateriaAttach
0x140225710: GetAgentTourismMenu
0x1402257F0: GetAgentPerformanceMode
0x1402258F0: GetAgentTofuList
0x140225940: GetAgent142
0x140225950: GetAgent433
0x140225980: GetAgentContext
```

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -3887,6 +3887,46 @@ classes:
       0x14022B470: HideAgentIfActive
       0x14022B4C0: IsAgentActive
       0x14022B230: HideAll
+      0x140225660: GetAgentScenarioTree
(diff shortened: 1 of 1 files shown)


wolfcomp:
I would say just add them even if they have no callers. Just so we have more named when going through the decompilation

pohky:
not with their id for names, the IDs change

Haselnussbomber:
oh right, i'll leave the ones with id out

PR #855, opened 2024-04-22
Make it possible for AgentId and InfoProxyId enums to be exported

Haselnussbomber:
Currently CExporter exports `AgentModule.GetAgentByInternalID(uint)` and `InfoModule.GetInfoProxyById(uint)`.
By swapping helper and member functions, CExporter will respect the existence of the corresponding enums and exports them.

Then we will be able to read the names in IDA, instead of having to look up the ids every time:

GetAgentByInternalID

GetInfoProxyById

PR #856, opened 2024-04-22
Obsolete uint Agent/InfoProxy getters

Haselnussbomber:
~~Stacked PR~~, based on https://github.com/aers/FFXIVClientStructs/pull/855 to make it an optional change.

This PR marks `GetAgentByInternalID(uint)` and `GetInfoProxyById(uint)` obsolete and makes the enum variants the primary and only getters. Users can just cast to AgentId/InfoProxyId if an something is missing in the enum. 🤷‍♂️


Haselnussbomber:
Not a breaking PR btw.

PR #857, opened 2024-04-25
Update breaking change check action

Haselnussbomber:
I added action logic to remove the label when no breaking change was found anymore.

This is untested, but please give it a try!

### DIFF
--- a/.github/workflows/pr-breaking-change-check.yaml
+++ b/.github/workflows/pr-breaking-change-check.yaml
@@ -68,14 +68,30 @@ jobs:
           echo "No breaking changes detected"
           echo "breaking=false" >> $GITHUB_OUTPUT
         fi
-    - name: Add tag to PR
-      if: steps.breaking_changes.outputs.breaking == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-breaking-change-check.yaml
```diff
@@ -32,6 +32,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.sha }}
```
wolfcomp, line 36: This isn't needed, and you should always rebase to main when updating the pr.

wolfcomp (changes requested):

.github/workflows/pr-breaking-change-check.yaml
Haselnussbomber: Nobody is rebasing when they add a couple of commits to fix something while the main branch does main branch things. Sorry, but that's just not how things work.

PR #875, opened 2024-05-26
DT Interop Generator Rewrite

aers:
This PR makes no changes to the existing library/generators and is purely additive with the new generator not enabled yet. This was done to avoid conflicts with other breaking changes merges in the lead-up to DT. Once we've merged those, we can do the work to transition the generator.

### Structure

- InteropGenerator - contains the generator as well as analyzers for diagnostics
- InteropGenerator.Runtime - contains the generator's attributes as well as the signature resolver
- InteropGenerator.Tests - contains unit tests for the generator and analyzers

The new generator replaces the generators that are not FFXIV-specific (MemberFunction, VirtualFunction, etc) and has been moved to its own namespace rather than being contained in FFXIVClientStructs.

While the old generators emit diagnostics directly, this is considered a bad practice by the C# language team. Emitting diagnostics in analyzers means the IDE can properly display errors at locations, the same way any other warning or error is displayed. The current list of errors can be found [here](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator/AnalyzerReleases.Unshipped.md).

### Features/Attributes

The specific code-gen for any attribute can be seen in the [generator tests](https://github.com/aers/FFXIVClientStructs/tree/dt-rewrite/InteropGenerator.Tests/Generator).

#### [GenerateInteropAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/GenerateInteropAttribute.cs)

New attribute. Must be used on any struct that you want the generator to touch. Since the generator is combined now, we can't trigger off individual method/field attributes like we used to, so this is necessary for performance reasons.

Has an optional argument, isInherited. This must be set to true if you intend for a struct to be inherited by another struct, but shouldn't be used otherwise. It isn't needed for a struct that inherits, only a struct that is inherited.

#### [MemberFunctionAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/MemberFunctionAttribute.cs)

Attribute usage mostly unchanged. There's a new optional argument, relativeFollowOffset, that works the same way as the offset in the StaticAddress attribute. To maintain the way it currently works, the generator automatically uses 1 for this argument if the signature starts with E8 or E9.

The larger change is in the codegen, which now looks like this:

```cs
public partial int TestFunction(int argOne, void* argTwo)
{
    if (MemberFunctionPointers.TestFunction is null)
    {
        InteropGenerator.Runtime.ThrowHelper.ThrowNullAddress("TestStruct.TestFunction", "AA BB CC DD ?? ?? ?? ?? AA BB ?? DD");
    }
    return MemberFunctionPointers.TestFunction((TestStruct*)global::System.Runtime.CompilerServices.Unsafe.AsPointer(ref this), argOne, argTwo);
}
```

Calling into a method that throws the exception instead of throwing directly allows better optimization The JIT disables some optimizations on methods that throw, and crucially methods that throw can never be inlined. Unsafe.AsPointer simply removes the use of fixed to get a pointer to this, which never made any sense when the objects were unmanaged.

#### [VirtualFunctionAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/VirtualFunctionAttribute.cs)

Attribute usage unchanged.

Now generates an accessor for the class's VirtualTable at offset 0 with the name VirtualTable. This will replace all the instances of void* vtbl/void** vfunc we have scattered throughout the codebase. If you still need a void** pointer for your own reasons you can cast it.

Renaming "vtbl" to "VirtualTable" will cause breaking changes for some users.

#### [VirtualTableAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/VirtualTableAttribute.cs)

Renamed from VTableAddress. Allows you to provide a signature for the address of the primary VirtualTable of a class. Generates a StaticVirtualTable field to access this address.

Will also generate the instanced VirtualTable struct/accessor if there aren't any instances of VirtualFunctionAttribute already causing it to happen.

#### [StaticAddressAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/StaticAddressAttribute.cs)

Unchanged. Has the same codegen improvements as MemberFunctionAttribute.

#### [GenerateStringOverloadsAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/GenerateStringOverloadsAttribute.cs) and [StringIgnoreAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/StringIgnoreAttribute.cs)

Renamed CStrOverloads attribute. 

No longer takes an argument for a parameter to be ignored; instead, add [StringIgnore] to byte* parameters that aren't strings. 

#### [FixedSizeArrayAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/FixedSizeArrayAttribute.cs)

Significant changes to this one. Will cause breaking changes due to renamed fields + using it in more places.

No longer takes type and size arguments or uses a fixed byte array field. Instead, the field must be formatted like follows:

```cs
[FieldOffset(0)] [FixedSizeArray] internal FixedSizeArray10<int> _arrayOfTenInts;
```

The field needs to be internal, have a name starting with _, and the type must be "FixedSizeArray#<T>" where # is the size and T is the contained type. Use Pointer<T> for pointers, as C# doesn't support pointer generics.

This will generate an accessor:

```cs
// <inheritdoc cref="_tenIntArray" />
[global::System.Diagnostics.CodeAnalysis.UnscopedRefAttribute] public Span<int> ArrayOfTenInts => _arrayOfTenInts;
```                           

Note that C# InlineArrays automatically convert to spans as a runtime feature.

And if necessary, creates the .NET 8.0 InlineArray type backing it:

```cs
[InlineArray(10)]
public struct FixedSizeArray10<T> where T : unmanaged {
    private T element0;
}
```

From the user perspective you will simply have a Span<T> to do whatever you want with, which is what we provide now. The backing field is made internal to stop it being used by consumers, since updating the size is now a breaking change if you use the type directly. The main goal here is to remove all the fixed byte array fields that are actually a different type, working around a C# limitation.

If you need access to the old fixed pointer, `Unsafe.AsPointer<T>(ref TheSpan[0])` should give you a pointer to the first element.

#### [InheritsAttribute](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Runtime/Attributes/InheritsAttribute.cs)

Allows a struct to inherit another struct, faking C++ polymorphism that isn't supported easily in C# interop scenarios.

```cs
[GenerateInterop(isInherited: true)]
public struct Base { ... }

[GenerateInterop]
[Inherits<Base>]
public struct Child { ... }
```

There is an optional parentOffset argument to InheritsAttribute that allows you to specify explicit offsets if a parent struct is missing in multiple inheritance scenarios, but most of the time the offsets are calculated using the size of parent structs.

The generator will copy:

- public fields
- FixedSizeArray accessors (but not the base internal field)
- MemberFunction 
- VirtualFunction
- entries for the primary virtual table
- other public methods that aren't generation targets
- string overloads

Static methods/addresses aren't copied because those aren't inherited.

The generator will also generate the accessor fields for parent classes (eg `[FieldOffset(0)] public AtkResNode AtkResNode;` and the explicit fields will get removed 

For specifics on codegen you can look at the [tests](https://github.com/aers/FFXIVClientStructs/blob/dt-rewrite/InteropGenerator.Tests/Generator/InheritsAttributeTests.cs). Most methods are copied by generating a stub that calls into the parent class via this accessor, but virtual functions in the primary virtual table call the native function directly.

The Inheritance feature has the least analyzer errors because doing analysis on inheritance trees isn't something I want to bother with. Don't create a circle :)

### I Hate Change

sorry

We wont be merging this PR immediately. I'm looking for feedback and feature suggestions. There are no stupid questions.


NotNite:
> The new generator replaces the generators that are not FFXIV-specific (MemberFunction, VirtualFunction, etc) and has been moved to its own namespace rather than being contained in FFXIVClientStructs.

I've been told in the XIVLauncher & Dalamud Discord server that this will allow usage from other projects. I'm very interested in this (particularly for other games that I want to bring CS-like development practices into). A few questions:

- How will configuring it work? I think I heard MSBuild properties?
- Is it possible to publish the source generator on NuGet? I've heard you technically can but then need to manually reference the .dll, which seems jank.
- I presume I can swap out the runtime project with my own (e.g. if I wanted my own signature engine), but then I'd need to keep the attributes it uses. Could/will there be the ability to specify a custom resolver class for the generator to use?

Thanks for all of the effort you've put into this so far!

aers:
It's possible to separate the project and publish to NuGet but that's future work since its slightly annoying to setup source generator stuff.

The only configuration is to set the namespace the generator puts some static stuff into - the FixedSizeArray structs and the resolver initializer function. The default namespace works if you only use the generator once in a project, but it will conflict if say a plugin references CS and uses the generator itself. Its configured through a MSBuild project property or in globalconfig (which can set MSBuild properties)

The generator currently preprocesses signatures and outputs them in a format specific to the resolver for performance so its pretty tightly coupled and will throw analyzer errors if the signature format doesn't match so its kinda hard to make it a generic interface as things stand..

PR #881, opened 2024-05-30
Enable deterministic CI builds

Haselnussbomber:
This PR enables

- [ContinuousIntegrationBuild](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild) to normalize file paths in pdb files, and
- [Deterministic](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation#deterministic) to produce byte-for-byte identical dlls across compilations for identical inputs.

### DIFF
--- a/FFXIVClientStructs/FFXIVClientStructs.csproj
+++ b/FFXIVClientStructs/FFXIVClientStructs.csproj
@@ -30,7 +30,12 @@
     
     <InteropGenerator_InteropNamespace>FFXIVClientStructs.Interop.Generated</InteropGenerator_InteropNamespace>
   </PropertyGroup>
-  
+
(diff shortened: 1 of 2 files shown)


wolfcomp:
Looking at the CIB flag and I noticed there was a note for it. Have you tested if this is an issue with the way the project is currently configured? https://github.com/dotnet/roslyn/issues/55860

Haselnussbomber:
I just briefly checked the paths in the pdb using a hex editor and they were normalized (`/_/FFXIVClientStructs/...`).
But the Runtime csproj should also have this defined. I'll add this really quick.

PR #888, opened 2024-06-02
Another cleanup

Haselnussbomber:
A bunch of code cleanup.

- added name/inheritance comments to most structs
- moved InheritsAttribute into it's own line
- moved VirtualTableAttribute above StructLayoutAttribute
- fixed wrong indentation caused by find and replace all
- ordered Instance() getters first, then fields, then methods, then structs and enums
- moved structs
  - `InfoProxyInvitedInterface` and `InfoProxyPageInterface` into their own files
  - `EventSceneModuleTaskManager` to its own file
  - `Status` struct from their own file into/after `Client::Game::StatusManager`
- added inheritance to
  - HousingOutdoorTerritory, which inherits from HousingTerritory
  - `Client::System::Input::ClipBoard`, which correctly inherits now from `Client::System::Input::ClipBoardInterface` (moved vfuncs too)
  - WorkDefinitions all inherit from `Application::Network::WorkDefinitions::Base` now
- renamed agents based on csv:
  - AgentFriendList -> AgentFriendlist
  - AgentHudLayout -> AgentHUDLayout
- AgentId renames:
  - SocialBlacklist -> Blacklist
  - SocialFriendList -> Friendlist
  - SocialPartyMember -> PartyMember
  - SocialSearch -> Search

---

Issues found:
- I let `Application::Network::WorkDefinitions::EnterContentInfo` inherit from `Application::Network::WorkDefinitions::Base`, but I'm not sure if that is correct. It's not in the data.yml and had a field at 0x00.
- All these fields in `Client::UI::UIInputData` belong to another struct `Component::GUI::AtkInputData` (not done in this PR). I just added `[Inherits<UserFileEvent>(0x9C8)]` to it, but not sure if that's right.


Haselnussbomber:
I assume these fields are part of `HousingTerritory` then?!

wolfcomp:
my best guess is that the actual outdoor and indoor territories are fully separate structs, but i couldnt fix it at the time since API breaking at the time

wolfcomp:
all 3 files were at one point documented in a single file but later split up into seperate ones but outdoor and indoor never got fully split appart.

My main point to why i argument for them actually being fully seperate structs are the fact that the workshop is a completely different struct compared to those two

Haselnussbomber:
Yep, they are separate structs, sharing an undocumented Interface at 141A77E98.
I'll see what I can do.

qPL2KMzNsv

Haselnussbomber:
There should be shared struct later which is used by IndoorTerritory and OutdoorTerritory, containing the Furniture array, the HousingObjectManager and more. They both call `sub_140B2B100` to set it up. I don't want to deal with this now.

PR #895, opened 2024-06-03
Replace some void pointers

Haselnussbomber:
Some void* fixes.


Haselnussbomber:
[Component/Exd/Sheets/TerritoryIntendedUse.cs](https://github.com/Haselnussbomber/FFXIVClientStructs/blob/7cfbdb0882c8a13f4454eb2884db8fccb447bedd/FFXIVClientStructs/FFXIV/Component/Exd/Sheets/TerritoryIntendedUse.cs) might also be a good place to put that enum (@MidoriKami). 😉

PR #919, opened 2024-06-14
[Dawntrail] Update Client::Game::InventoryItem

Haselnussbomber:
For the data.yml, a new SetStain function was added at `E8 ?? ?? ?? ?? 0F B6 53 1D`


Haselnussbomber:
Just for the record on why this PR history looks so weird: wildwolf asked me to test some GitHub Actions stuff, which removes the comment and label if it's no longer having issues.
I did a couple of fake commits to make it happy and then reverted it again. I won't be changing other structs sizes or offsets in here.

PR #924, opened 2024-06-15
[Dawntrail] Update offsets

Nukoooo:


FFXIVClientStructs/FFXIV/Client/Game/Object/GameObjectManager.cs
pohky, line 1: this needs to be reverted to what it was
GameObjectManager can't be updated based on the benchmark as they use completely different sizes/object counts for release and benchmark

Nukoooo: Done

PR #928, opened 2024-06-20
Add AtkValue.ToString() overload

Haselnussbomber:
Based on helper from @MidoriKami: https://discord.com/channels/581875019861328007/1246473943548825730/1253413305230557290

Untested. Please have a look!


MidoriKami:
Based on helper from Caraxi =D

PR #932, opened 2024-06-21
Add AtkListItemData and fix entirely unrelated struct size

MidoriKami:


Haselnussbomber:
As a note:
1) We change the size of `AtkEventData` to 0x28, because it copies the data inside `AtkEventDispatcher_DispatchEvent` using OWORDs, not QWORDs.
2) `AtkListItemData` is prepared and dispatched inside `AtkComponentList_vf44`.

PR #936, opened 2024-06-23
Add layout prefetch functions to LayoutWorld

Haselnussbomber:
Adds a function to prefetch a layout and another one to cancel it.

`LayoutWorld.PrefetchLayout` is called with type 1 and [LayerEntryType](https://github.com/NotAdam/Lumina/blob/master/src/Lumina/Data/Parsing/Layer/LayerEnums.cs) 71 (PrefetchRange) when near a zone barrier, or with type 2 and LayerEntryType 40 (PopRange) when starting to cast Teleport.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/LayoutEngine/LayoutWorld.cs
+++ b/FFXIVClientStructs/FFXIV/Client/LayoutEngine/LayoutWorld.cs
@@ -1,3 +1,4 @@
+using FFXIVClientStructs.FFXIV.Client.Game;
 using FFXIVClientStructs.FFXIV.Client.System.String;
 
 namespace FFXIVClientStructs.FFXIV.Client.LayoutEngine;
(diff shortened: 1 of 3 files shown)


Haselnussbomber:
Oh right, I changed the function name last minute. Now I’m in bed already. Will fix it when I wake up again. 😴

FFXIVClientStructs/FFXIV/Client/LayoutEngine/LayoutWorld.cs
```diff
@@ -18,12 +19,18 @@ public unsafe partial struct LayoutWorld {
     [FieldOffset(0x018)] public LayoutManager* GlobalLayout;
     [FieldOffset(0x020)] public LayoutManager* ActiveLayout;
     //[FieldOffset(0x028)] public LayoutManager* UnkLayout28;
-    //[FieldOffset(0x030)] public LayoutManager* UnkLayout30;
+    [FieldOffset(0x030)] public LayoutManager* PrefetchLayout;
```
wolfcomp, line 22: If this contains the pointer to the prefetched layout, it should be named as such to not have a collision in IDA

PR #958, opened 2024-06-29
Emit generated source files

goaaats:
This is required for docs generation on our end. docfx can't work directly with source generators.
If it was disabled on purpose, maybe it could be enabled conditionally?


wolfcomp:
It was disabled because didn't want to commit them to the git if I recall. better to wrap it in a condition that can be toggled on if it's generated under docfx.

goaaats:
I'll override it from nuke, might be better like that anywho.

PR #1007, opened 2024-07-03
Update Client::Game::UI::MobHunt

Orphis:
Fixes the offset and array sizes for the 4 new Dawntrail MobHunt order types.
Adds a constant MaxMarkIndex that can be used to refer to 1 past the last MobHuntOrderType index, which is returned by `GetMarkIndexFromItemId(uint)` or to iterate over all the order types.


MidoriKami:
This looks exactly like what I came up with, but I was having other troubles verifying, thanks for PRing this.

PR #1008, opened 2024-07-03
idarename, structimporter: Add Binary Ninja support

NotNite:
I doubt anyone can test this but it seems to work on my machine. Will eventually add structimporter support when I have the motivation.


NotNite:
Added structimporter as well; I'll probably get around to exdgetters later.

image

PR #1011, opened 2024-07-04
Minor additions to ActionManager, PlayerState and LimitBreakController.

awgil:


FFXIVClientStructs/FFXIV/Client/Game/ActionManager.cs
```diff
@@ -267,11 +267,16 @@ public unsafe partial struct ActionManager {
...
+    /// Determine character's category for action targeting purposes; this is used by the game to determine whether a spell can be used on that target.
+    /// </summary>
+    [MemberFunction("40 53 48 83 EC 20 48 8B 01 48 8B D9 FF 50 08 8B 15")]
+    public static partial TargetCategory ClassifyTarget(Character.Character* target);
+
     public enum CastTimeProc : byte {
         None = 0,
         Firestarter = 1, // THM/BLM
-        Thundercloud = 2, // THM/BLM
+        Thundercloud = 2, // THM/BLM, gone in 7.0
```
pohky, line 279: if it's gone just comment the whole line out for now, no need to have the value named wrong

awgil: Done.
I don't think it's actually deleted in SE's source (since remaining enums didn't shift, and i doubt they assign explicit indices), but it's never returned anymore. Not a big deal either way.

PR #1022, opened 2024-07-08
Small changes to NamePlateObjectIntArrayData

nebel:


nebel:
This is breaking but since it's just a week old and I don't think anyone's using it yet, hopefully it's fine.

PR #1028, opened 2024-07-10
Add SetGlasses to DrawDataContainer

chirpxiv:


FFXIVClientStructs/FFXIV/Client/Game/Character/DrawDataContainer.cs
```diff
@@ -58,6 +58,14 @@ public unsafe partial struct DrawDataContainer {
     [MemberFunction("E8 ?? ?? ?? ?? 0F B6 97 ?? ?? ?? ?? 48 8B CF C0 EA")]
     public partial void SetVisor(bool state);
 
+    /// <summary>
+    /// Called when equipping facewear.
+    /// </summary>
+    /// <param name="unk">Usually 0.</param>
+    /// <param name="id">Row ID from the Glasses sheet.</param>
+    [MemberFunction("E8 ?? ?? ?? ?? EB 50 44 8B 03")]
+    public partial void SetGlasses(uint unk, ushort id);
```
Haselnussbomber, line 67: There are 2 slots for whatever reason, so this parameter is the index.

```suggestion
    /// <param name="index">The index of the glasses slot, usually 0.</param>
    /// <param name="id">Row ID from the Glasses sheet.</param>
    [MemberFunction("E8 ?? ?? ?? ?? EB 50 44 8B 03")]
    public partial void SetGlasses(int index, ushort id);
```

Haselnussbomber (changes requested):

PR #1044, opened 2024-07-15
Fix AddonNeedGreed

MidoriKami:


MidoriKami:
Hmmm, that doesn't seem right.

Haselnussbomber:
Your branch is 12 commits behind aers/FFXIVClientStructs:main, which is why all this stuff is being detected as breaking change.

MidoriKami:
Ah, right I remember that coming up before. 👍

PR #1049, opened 2024-07-16
Tolerate incomplete exd schema.

awgil:


awgil:
Is a slight schema mismatch a good reason to fail the entire exdgetters, though?

PR #1059, opened 2024-07-19
Update CharacterUtility for 7.0

Exter-N:


wolfcomp:
I can't see where you found that 114 number, but I do see it is higher now, so guess it's just in a function that isn't part of the vtable

PR #1065, opened 2024-07-22
Better Texture Width/Height/Format

Exter-N:
- Attempts to give better names to `Width`, `Height`, `Width2` and `Height2`, following observations of cases where they differ ;
- Describes more `TextureFormat` values, according to `FUN_1403f0700` and [DXGI_FORMAT](https://learn.microsoft.com/en-us/windows/win32/api/dxgiformat/ne-dxgiformat-dxgi_format).


wolfcomp:
Lumina gets their data from the `.dat` files and maps the IDs from there. It could very much be that they are converting said data after the fact in the engine.

Soreepeong:
unorm 2/3: the game converts 5551 and 4444 to 8888 upon reading the texture before uploading to gpu.

FFXIVClientStructs/FFXIV/Client/Graphics/Kernel/Texture.cs
```diff
@@ -48,8 +54,49 @@ public unsafe partial struct Texture {
     public partial void DecRef();
 }
 
-// there's 20+ of these but these are the ones I've encountered/debugged
+// See also DXGI_FORMATs that end with the same names.
 public enum TextureFormat : uint {
-    R8G8B8A8 = 5200,
-    D24S8 = 16976 // depth 28 stencil 8, see MS texture formats on google if you really care :)
+    B8G8R8A8_UNORM_4 = 0x1130,
```
wolfcomp, line 59: Probably named `L8_UNORM` but this is guessing based on what Lumina has documented

wolfcomp (changes requested):
Just a few changes for it to be good.

PR #1076, opened 2024-07-26
Agent content finder

Infiziert90:
Adjusted with @wolfcomp's help

`AgentContentsFinderUnkItemsSub` seems to be yet another copy of all the items


wolfcomp:
Checked [GitHub](https://github.com/search?q=AgentContentsFinder+language%3AC%23+NOT+is%3Afork&type=code) and see 0 uses of the breaking changes listed. But as this is also the correct structure for the game files, it's ok to be merged even as it is labeled as breaking.

PR #1094, opened 2024-08-16
Update RaptureLogModule.GetLogMessageDetail

Haselnussbomber:
The logKind parameter of the existing GetLogMessageDetail function is incorrect.
It contains the LogKind RowId in the first 7 bits, then 4 bits of casterKind and 4 bits of targetKind.
I've added a new overload that splits this information into new out params and obsoleted the old one.


wolfcomp:
Can you send the Func you found this in?

Haselnussbomber:
I found it almost two weeks ago, so I don't exactly remember where it was. Somewhere where it adds the message to the array.
You can see it masking the logKind with 0x7F after the only call of the function.
It does the same masking and shifting here 1407B41E2 to pass it to FormatLogMessage.
The addon texts come from 140300A60 (which is inlined in another function that ida can't decompile).

PR #1096, opened 2024-08-17
Fix return type of RaptureGearsetModule.UpdateGearset

Haselnussbomber:
The function returns error codes (-1 or -2) or the gearset id if successful.
If the function returns anything under -1, the gearset list will refresh and scroll up (which is the issue I'm facing right now).


Haselnussbomber:
Moved to #1089

PR #913, opened 2024-06-11
Update AtkUnitBase

Haselnussbomber:
Going through AtkUnitBase again, adding flag fields as they are used by the game and doing some cleanup.

Along with the flags, a couple notable changes:
- Added `AtkResNode.IsVisible` function
  It looks like someone found a function to check if an AtkResNode is visible. It does a little bit more than just checking the flag, so I included it. This obviously changes it from being a property to being a function.
- Added `AtkUnitBase.IsFullyLoaded` (vf65) and removed `AtkUldManager.SearchNodeById` wrapper
  This virtual function checks if all resources (uld and optional scd) are loaded and if the UldManager loaded the uld and all the textures it needs. This function should be the preferred way of checking if an addon was fully loaded going forward.
- Added `AtkExternalInterface` vfuncs 1 (CallHandler) and 2 (PlaySoundEffect).

PR #1098, opened 2024-08-18
Add ICreatable to Components

MidoriKami:
AtkUldManager function was added in https://github.com/aers/FFXIVClientStructs/pull/913 but got nuked during a merge conflict.

Added ICreatable to all Components that I could find


MidoriKami:
Used VirtualTable attribute where practical, sigged orphaned ctors otherwise

PR #1107, opened 2024-08-24
Fixed cexporter offsets for non-explicit-layout structs.

awgil:
This is important to fix e.g. std::map / std::pair structures


wolfcomp:
Different method implmeneted

PR #1109, opened 2024-08-27
Add AgentActionDetail

Infiziert90:


Infiziert90:
`140efb910` resolves all ActionKinds so it was easy to map them

PR #1112, opened 2024-09-02
Random bits mostly related to framework and gameobject/character.

awgil:


awgil:
I think bot can be ignored here. GameObject.GetRadius() function now indeed takes a (bool) argument (you can see that eg. in Character.GetRadius override), but having a default should make the change non-breaking for users.

PR #1121, opened 2024-09-23
Update UIState

Haselnussbomber:
- Fix some struct sizes that weren't updated with 7.0
- Add some random fields
- Moved `Instance()` function above fields for visibility, like it is in other structs

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/UI/UIState.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/UI/UIState.cs
@@ -1,5 +1,4 @@
-using FFXIVClientStructs.FFXIV.Client.Game.Event;
-using FFXIVClientStructs.FFXIV.Client.Game.Fate;
+using FFXIVClientStructs.FFXIV.Application.Network.WorkDefinitions;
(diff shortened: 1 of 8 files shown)


FFXIVClientStructs/FFXIV/Client/Game/UI/UIState.cs
```diff
@@ -66,7 +69,10 @@ public unsafe partial struct UIState {
     [FieldOffset(0x16398)] public GoldSaucerYell GoldSaucerYell;
     [FieldOffset(0x17AE8)] public CharaCard CharaCard;
     // 0x17CD0: ItemAction Unlocks
-
+    [FieldOffset(0x17D28)] public ushort CharaViewDrawDataFlags; // & 1 = HideHeadgear, & 2 = HideWeapons, & 4 = HideLegacyTattoo, & 0x40 = HideVisor
```
wolfcomp, line 72: Should this be made to an enum flag type?

Haselnussbomber: Found out what it is. :)

PR #1132, opened 2024-10-18
Add UIGlobals struct

Haselnussbomber:
This PR adds the `UIGlobals` struct, which serves as a central place for storing global static functions from the `Client::UI` namespace.

Some of these global functions were added to other structs in the past, which is why they have been moved to UIGlobals:
- `GameMain.IsInSanctuary`, moved as `CanApplyGlamourPlates` (see also: #1123)
- `UIModule.PlaySound`, moved as `PlaySoundEffect`
- `UIModule.IsPlayerCharacterName`, moved as `IsValidPlayerCharacterName`
- `UIModule.PlayChatSoundEffect`

The entries in the data.yml have also been renamed accordingly, though I did not add the namespace prefix if it was missing.

I've also added the `IsExportedGatheringPointTimed` function (formerly named IsGatheringTypeRare by me in the data.yml) to this struct (because I need it myself), and tried to add some xmldocs to these functions.

While I did not dig into the `PlaySoundEffect` function, which is why I didn't add a comment there, I changed the unknown argument types from `long` to `nint`, because from what I could see both are pointers: a2 is an out parameter for a string and a3 is an out parameter for some kind of instance of a sound effect struct. 🤷‍♂️

Since this is a new struct and the old functions have been obsoleted, there are no breaking changes in this PR.
Please let me know if I did something wrong.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/UIGlobals.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/UIGlobals.cs
@@ -1,0 +1,54 @@
+namespace FFXIVClientStructs.FFXIV.Client.UI;
+
+/// <summary>
+/// This struct is a collection of static functions from the Client::UI namespace.
+/// </summary>
+[GenerateInterop]
+public unsafe partial struct UIGlobals {
+    /// <summary>
+    /// Determines whether the player can apply glamour plates in the current territory.
+    /// </summary>
+    /// <param name="checkCastGlamourUnlocked">
+    /// If <c>false</c>, skips the check for whether the player has unlocked the general action "Cast Glamour".
+    /// </param>
+    /// <returns>
+    /// <c>true</c> if the player can apply glamour plates in the current territory; otherwise, <c>false</c>.
+    /// </returns>
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 75 21 48 8B 4F 10")]
+    public static partial bool CanApplyGlamourPlates(bool checkCastGlamourUnlocked = true);
+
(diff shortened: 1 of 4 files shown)


FFXIVClientStructs/FFXIV/Client/UI/UIGlobals.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    /// <param name="characterName">The character name of the player to validate.</param>
+    /// <returns>
+    /// <c>true</c> if the character name is valid; otherwise, <c>false</c>.
+    /// </returns>
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 74 B4 4C 8B CB"), GenerateStringOverloads]
+    public static partial bool IsValidPlayerCharacterName(byte* characterName);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 63 45 80")]
+    public static partial void PlaySoundEffect(uint effectId, nint a2 = 0, nint a3 = 0, byte a4 = 0);
```
wolfcomp, line 46: If `AtkValue.Int64` **CAN NOT** be pointer, this is the correct way of invoking the function.
ref: `Component::GUI::AtkModule_PlaySoundEffectHandler`
```suggestion
    public static partial void PlaySoundEffect(uint effectId, long a2 = 0, long a3 = 0, byte a4 = 0);
```

wolfcomp: #1137 has confirmed it is pointer type

PR #1150, opened 2024-11-10
Add LinkData (for AtkTextNode/AtkEventData)

Haselnussbomber:
Unsure how this all fits together. There is a lot of font stuff involved (to calculate text positions?).

Also, I didn't know how to properly map PayloadEndAndLinkType, so I added CExportIgnoreAttribute to it.
- At `41 8B 4C 24 ?? 41 8B 5C 24` it uses a bitshift to get the payload length:
  `v5 = (int)(*(_DWORD *)&v2->PayloadEndAndLinkMacroPayloadType << 8) >> 8;`
- At other places (e.g. `41 0F BE 7D ?? 83 FF 03`) it gets the `LinkType` directly.

Also, UIntValue1 to 3 are the link parameters, but maybe they should be Int?

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
@@ -143,12 +143,12 @@ public unsafe struct LinkData {
     [FieldOffset(0x1E)] public ushort Unk1E;
     [FieldOffset(0x20)] public uint Unk20;
     // These are the 3 link payload parameters. Usually SeStrings have int expressions.
-    [FieldOffset(0x24)] public int IntValue1;
-    [FieldOffset(0x24), CExportIgnore] public uint UIntValue1;
-    [FieldOffset(0x28)] public int IntValue2;
-    [FieldOffset(0x28), CExportIgnore] public uint UIntValue2;
-    [FieldOffset(0x2C)] public int IntValue3;
-    [FieldOffset(0x2C), CExportIgnore] public uint UIntValue3;
+    [FieldOffset(0x24), CExporterUnion("Value1")] public int IntValue1;
+    [FieldOffset(0x24), CExporterUnion("Value1")] public uint UIntValue1;
+    [FieldOffset(0x28), CExporterUnion("Value2")] public int IntValue2;
+    [FieldOffset(0x28), CExporterUnion("Value2")] public uint UIntValue2;
+    [FieldOffset(0x2C), CExporterUnion("Value3")] public int IntValue3;
+    [FieldOffset(0x2C), CExporterUnion("Value3")] public uint UIntValue3;
     [FieldOffset(0x30)] public ulong UnkAndBackgroundColor;
 
     // length from the start of the text in AtkTextNode or so. not sure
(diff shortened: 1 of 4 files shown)


FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
```diff
@@ -128,3 +126,33 @@ public enum FontType : byte {
...
+    [FieldOffset(0x1C)] public ushort Unk1C;
+    [FieldOffset(0x1E)] public ushort Unk1E;
+    [FieldOffset(0x20)] public uint Unk20;
+    // These are the 3 link payload parameters. Usually SeStrings have int expressions.
+    [FieldOffset(0x24)] public int IntValue1;
+    [FieldOffset(0x24), CExportIgnore] public uint UIntValue1;
+    [FieldOffset(0x28)] public int IntValue2;
+    [FieldOffset(0x28), CExportIgnore] public uint UIntValue2;
+    [FieldOffset(0x2C)] public int IntValue3;
+    [FieldOffset(0x2C), CExportIgnore] public uint UIntValue3;
```
wolfcomp, line 151:
```suggestion
    [FieldOffset(0x24), CExporterUnion("Value1")] public int IntValue1;
    [FieldOffset(0x24), CExporterUnion("Value1")] public uint UIntValue1;
    [FieldOffset(0x28), CExporterUnion("Value2")] public int IntValue2;
    [FieldOffset(0x28), CExporterUnion("Value2")] public uint UIntValue2;
    [FieldOffset(0x2C), CExporterUnion("Value3")] public int IntValue3;
    [FieldOffset(0x2C), CExporterUnion("Value3")] public uint UIntValue3;
```

Use that if it depends which one it uses based on int or uint.

Another case could be that they are doing type casting in the code and there is only one value type

FFXIVClientStructs/FFXIV/Component/GUI/AtkTextNode.cs
```diff
@@ -128,3 +126,33 @@ public enum FontType : byte {
...
+[StructLayout(LayoutKind.Explicit, Size = 0x38)]
+public unsafe struct LinkData {
+    [FieldOffset(0x00)] public int Unk0;
+    [FieldOffset(0x04)] public int Unk4;
+    [FieldOffset(0x08)] public int Unk8;
+    [FieldOffset(0x0C)] public int UnkC;
+    /// <remarks> To be read until link terminator. </remarks>
+    [FieldOffset(0x10)] public byte* Payload;
+    /// <remarks> First 3 bytes are PayloadEnd(?) and last byte is <see cref="LinkType"/> </remarks>
+    [FieldOffset(0x18), CExportIgnore] public uint PayloadEndAndLinkType;
```
wolfcomp, line 139: Is this offset set as an unsigned int somewhere?

Haselnussbomber: Well, it contains 2 values, so it wouldn't make sense if it's int, because one bit is used for the sign then.
It converts it after bitshifting to an int, but it should be uint before that, or you get wrong values, I think?
I wrote the sigs in the original message where you can find it.

Haselnussbomber: And coverting it to an int after bitshifting also doesn't make sense, becaue you'd have the sign at the wrong position. But that's what IDA shows. Not sure if accuate.

wolfcomp:
For later:
`AtkInputManager->field_1898` some char flag
`AtkInputManager->field_1890` some x value
`AtkInputManager->field_1894` some y value
func pre 7.1: 1406663D0

PR #1173, opened 2024-11-14
Update HousingManager sig

Limiana:


Haselnussbomber:
Oops, looks like I picked the sig for `HasInstance`. 😅 Thanks!

PR #1175, opened 2024-11-14
Update AgentHUD.cs

Aireil:
I couldn't check the size of HudQueuedBattleTalk.


Haselnussbomber:
`ExpCurrentExperience`, `ExpNeededExperience`, `ExpRestedExperience` and `ExpClassJobId` should be `int` instead of `uint` (my bad).
I just saw this when I was in the title screen:
Screenshot

PR #1194, opened 2024-11-16
Update GetEntry sigs in InfoProxyCommonList.cs

Aireil:


Aireil:
The second sig is wrong.

PR #1205, opened 2024-11-17
Update CharacterBase offsets.

Ottermandias:


aers:
i'll fix the class size

PR #1218, opened 2024-11-20
Update AgentItemSearch

Haselnussbomber:
It got bigger by 8 bytes due to the InventoryItem vtable. I also added some new fields.


Haselnussbomber:
I think ListingItem.ItemId is wrong. Let me have a second look.

PR #1222, opened 2024-11-22
Update Character.cs

Ottermandias:


aers:
Should this be an error obsolete? I assume the function call just doesn't work at all since it will pass chara as a1.

PR #1243, opened 2024-12-07
structimporter: use LibYAML loader if available

Haselnussbomber:
As per https://pyyaml.org/wiki/PyYAMLDocumentation, we can make use of LibYAML, which is a python parser written in c that is much faster than the default one written in python.
Worked out of the box for me and cut down ffxiv_structs.yml loading time from 29.4s to 2.8s.

### DIFF
--- a/ida/ffxiv_structimporter.py
+++ b/ida/ffxiv_structimporter.py
@@ -2,7 +2,12 @@
 # @menupath Tools.Scripts.ffxiv_structimport
 # @runtime Jython
 
-from yaml import safe_load
+from yaml import load
+try:
+    from yaml import CSafeLoader as Loader
+except ImportError:
+    from yaml import SafeLoader as Loader
+
 import os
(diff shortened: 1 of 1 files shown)


ida/ffxiv_structimporter.py
```diff
@@ -2,7 +2,12 @@
 # @menupath Tools.Scripts.ffxiv_structimport
 # @runtime Jython
 
-from yaml import safe_load
+from yaml import load
+try:
+    from yaml import CSafeLoader as Loader
+except ImportError:
+    from yaml import SafeLoader
```
wolfcomp, line 9:
```suggestion
    from yaml import SafeLoader as Loader
```

This will probably fail if this isn't here

Haselnussbomber: Oooooh right. I changed that afterwards because it wasn't the Safe one!

PR #1260, opened 2024-12-23
Fix PacketDispatcher statics

Haselnussbomber:
They are getting passed argument 2 (`uint targetId`) of `PacketDispatcher.OnReceivePacket`, not argument 1 (`PacketDispatcher* this`).
While none of these functions are using argument 1, it's still wrong.


wolfcomp:
Due to these being used in Dalamud it will be held off until next API update just for safety.

PR #1269, opened 2025-01-03
AgentCharaCard: Fix reading decorations

redstrate:
I'm a total idiot and only tested this code against a full adventurer plate, and it turns out almost everything I did was completely wrong. I assumed the location of each decoration is static, but it turns out this structure is more extendable than I thought, I guess to support more decoration types in the future.

The decoration data is stored in two different fixed size arrays with five elements (since there's only five decoration types.) The first array is the row indices, where the hard-coded locations were before. The second array is some more decoration information, but currently I only know of and care about the type for now.

This does not affect base plates or borders, which are not considered decorations.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentCharaCard.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentCharaCard.cs
@@ -25,22 +25,48 @@ public unsafe partial struct AgentCharaCard {
     private partial void OpenCharaCardForObject(GameObject* gameObject);
     public void OpenCharaCard(GameObject* gameObject) => OpenCharaCardForObject(gameObject);
 
+    public enum DecorationType
+    {
+        /// This does not correspond to a real decoration and should be ignored.
+        Invald = 0x0,
+        Backing = 0x1,
+        PatternOverlay = 0x2,
+        PortraitFrame = 0x3,
+        PlateFrame = 0x4,
+        Accent = 0x5,
+    }
+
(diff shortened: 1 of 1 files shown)


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentCharaCard.cs
```diff
@@ -68,6 +94,10 @@ public unsafe partial struct Storage {
         [FieldOffset(0x490)] public Utf8String Activity6Name;
 
         [FieldOffset(0x540)] public CharaViewPortrait CharaView;
+
+        /// The size of this array is NumDecorations.
+        [FieldOffset(0x22C), FixedSizeArray] internal FixedSizeArray5<DecorationSet> _decorations;
```
redstrate, line 99: I've seen most (if not all) fixed size arrays are marked internal, I guess I should expose something else in the API?

Haselnussbomber: The reason it's internal is to hide the `_decorations` field, because a `Span<DecorationSet> Decorations` is generated by the code generator.
You basically did everything right here, except that the type `DecorationSet` doesn't exist. (Did you mean `Decoration`?)

redstrate: Yeah copy & paste error, fixing that now :)

wolfcomp:
Will be held off until next API bump *possibly 7.2*

PR #1275, opened 2025-01-11
Update data.yml, Add AgentInventoryContext::InventoryContextEvent

Haselnussbomber:
Just for inheritance. Didn't look into what it does.

But I noticed that the ctor for `AgentAquariumSetting` sets it vtable to +8. Not sure what's going on there, doesn't make sense.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Event/ShopEventHandler.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Event/ShopEventHandler.cs
@@ -1,11 +1,15 @@
 using FFXIVClientStructs.FFXIV.Client.System.String;
+using FFXIVClientStructs.FFXIV.Client.UI.Agent;
 using FFXIVClientStructs.FFXIV.Common.Component.Excel;
 using FFXIVClientStructs.FFXIV.Component.GUI;
 
 namespace FFXIVClientStructs.FFXIV.Client.Game.Event;
 
+// Client::Game::Event::ShopEventHandler
+//   Client::Game::Event::EventHandler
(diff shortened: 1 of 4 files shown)


FFXIVClientStructs/FFXIV/Client/Game/Event/ShopEventHandler.cs
```diff
@@ -1,11 +1,15 @@
...
 using FFXIVClientStructs.FFXIV.Component.GUI;
 
 namespace FFXIVClientStructs.FFXIV.Client.Game.Event;
 
+// Client::Game::Event::ShopEventHandler
+//   Client::Game::Event::EventHandler
+//   Client::UI::Agent::AgentInventoryContext::InventoryContextEvent
 [GenerateInterop]
-[Inherits<EventHandler>]
+[Inherits<EventHandler>, Inherits<AgentInventoryContext.InventoryContextEvent>]
```
wolfcomp, line 12: This seems to have a 3rd inheritance that is calling stuff related to `Client::Game::Event::FinalizeSceneTask` 2 layers deep in vf1 of said inheritance. Not sure if you want to investigate this before merge.

Haselnussbomber: Yeah I know, says so in the comments at the start of the struct.
But I didn't want to look into it, because I didn't want to come up with a name. 😅

wolfcomp: `ShopSceneTaskScheduler` 😛

PR #1278, opened 2025-01-18
Update various thingies

Haselnussbomber:
Thought I'd PR before it gets even more.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInventoryContext.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInventoryContext.cs
@@ -49,7 +49,7 @@ public unsafe partial struct AgentInventoryContext {
     }
 
     [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 48 8D 0D ?? ?? ?? ?? E8 ?? ?? ?? ?? 4C 89 74 24 ??")]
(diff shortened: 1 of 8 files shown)


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInventoryContext.cs
```diff
@@ -49,7 +49,7 @@ public void OpenForItemSlot(InventoryType inventory, int slot, uint addonId) {
     }
 
     [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 48 8D 0D ?? ?? ?? ?? E8 ?? ?? ?? ?? 4C 89 74 24 ??")]
-    public partial long UseItem(uint itemId, uint inventoryType = 9999, uint itemSlot = 0, short a5 = 0);
+    public partial long UseItem(uint itemId, uint inventoryType = 9999, uint itemSlot = 0, short a5 = 0); // TODO: use InventoryType.Invalid
```
wolfcomp, line 52: Wouldn't it make more sense to update the `inventoryType` to the enum type, instead of just updating the default type to use the enum?

Haselnussbomber: I didn’t mean to write `(uint)InventoryType.Invalid`, lol.
Of course the type should be changed too, but I didn’t want to touch it right now.

PR #1287, opened 2025-01-23
Update Utf8String.SanitizeString, add GetParts

Haselnussbomber:
**Summary:**

- Added an `AllowedEntities` enum for the `Utf8String.SanitizeString` function.
- Added `Utf8String.GetParts`, which extracts the different texts and payloads from a string and adds them to a vector.

---

**Regarding SanitizeString:**

The commonly used flags 0x27F, which I don't know where those comes from, allow the following:
- UppercaseLetters
- LowercaseLetters
- Numbers
- SpecialCharacters
- CharacterList
- NewLines
- Payloads
- Unknown9 (Chinese/Korean client specific?)

---

**GetParts Example:**

```cs
var str = Utf8String.FromSequence(ReadOnlySeString.FromMacroString("<string(gstr1)><string(gstr2)><bold(1)>Hello @ World<bold(0)>"));
var vector = new StdVector<Utf8StringPart>();

str->GetParts(&vector);

foreach (var part in vector)
{
    ImGui.TextUnformatted($"Payload: {new ReadOnlySeStringSpan(part.Payload.AsSpan()).ToString()}");
    ImGui.TextUnformatted($"Text: {new ReadOnlySeStringSpan(part.Text.AsSpan()).ToString()}");
    ImGui.Separator();

    part.Payload.Dtor();
    part.Text.Dtor();
}
```

Renders:

Screenshot

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/System/String/Utf8String.cs
+++ b/FFXIVClientStructs/FFXIV/Client/System/String/Utf8String.cs
@@ -170,6 +170,9 @@ public unsafe partial struct Utf8String : ICreatable, IDisposable, IStaticNative
     [Obsolete("Use SanitizeString with AllowedEntities enum")]
     public void SanitizeString(ushort flags, Utf8String* characterList) => SanitizeString((AllowedEntities)flags, characterList);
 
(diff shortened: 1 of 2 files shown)


Haselnussbomber:
Renamed NewLines to Unicode, which is still incorrect, but eh..
It's basically everything(?) not handled by the alphanumeric filters:
Unicode

wolfcomp:
Could you check a bit more into what is chucked into the unicode parts?
For instance, is it all the latin characters that is region specific?

PR #1292, opened 2025-01-29
Fix CompletionModule

Haselnussbomber:
https://youtu.be/CD6o49RYEn4


Haselnussbomber:
This was all added an hour ago with #1290.

PR #1295, opened 2025-01-30
Implement IDA 9 support

salanth357:
Adds support for IDA 9
- abstracted a few helper methods out to an ida_helpers import so they can be reused between scripts

The script will fail on duplicate entries in vtables in ffxiv_structs.yml, but I'm surprised that isn't already an issue...
I _think_ these APIs should still work against IDA 7, but it'll need testing


wolfcomp:
image
Does not work on IDA 8 and 9 and thus will be blocked from being approved.

wolfcomp:
With 174 lines duplicated in `ffxiv_exdgetters.py` and 363 lines duplicated in `ffxiv_structimporter.py` this is not maintainable. As such it will not be approved for a solution to fixing IDA 9 support.

ida/ffxiv_exdgetters.py
```diff
@@ -344,7 +559,7 @@ def process_pattern(self, pattern):
                     self.do_pattern(pattern, suffix, True)
                 pass
 
-        api = IdaApi()
+        api = Ida7Api()
```
wolfcomp, line 562: Cut down on the duplicate code here

ida/ffxiv_structimporter.py
```diff
@@ -1002,8 +1525,7 @@ def should_update_virt_func(self):
             )
             == ida_kernwin.ASKBTN_YES
         )
-        api = IdaApi(full_padding)
-
+        api = Ida7Api(full_padding)
```
wolfcomp, line 1528: Cut down on the duplicate code here

wolfcomp (changes requested):

PR #1302, opened 2025-02-06
Switch to Conditions fields

Haselnussbomber:
Turns out not having named fields is pretty annoying while reverse engineering...

Other changes:

- Renamed a couple fields (the first 24 are based on CharacterMode):
  - NormalConditions -> Normal
  - Unconscious -> Dead
  - OperatingSiegeMachine -> AnimLock
  - CarryingObject -> Carrying
  - Mounted2 -> RidingPillion
- Obsoleted all static properties. I don't think it's worth keeping them.
- Added a float at the end, which is set in the `SetCondition` function.
- Dropped comments as LogMessage RowIds can be retrieved from the Condition sheet.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Conditions.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Conditions.cs
@@ -1,495 +1,704 @@
 namespace FFXIVClientStructs.FFXIV.Client.Game;
 
+// Client::Game::Condition
 [GenerateInterop]
-[StructLayout(LayoutKind.Explicit, Size = 104)]
+[StructLayout(LayoutKind.Explicit, Size = 104 + 4)]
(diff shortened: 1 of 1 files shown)


FFXIVClientStructs/FFXIV/Client/Game/Conditions.cs
```diff
@@ -1,495 +1,704 @@
 namespace FFXIVClientStructs.FFXIV.Client.Game;
 
+// Client::Game::Condition
 [GenerateInterop]
-[StructLayout(LayoutKind.Explicit, Size = 104)]
+[StructLayout(LayoutKind.Explicit, Size = 104 + 4)]
```
wolfcomp, line 5: why not just set size `108`?

PR #1303, opened 2025-02-08
Update RaptureAtkColorDataManager

Haselnussbomber:
For whoever needs it.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkColorDataManager.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkColorDataManager.cs
@@ -1,5 +1,22 @@
 namespace FFXIVClientStructs.FFXIV.Client.UI;
 
 // Client::UI::RaptureAtkColorDataManager
+// Used by the ColorPicker addon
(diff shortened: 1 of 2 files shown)


FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkColorDataManager.cs
```diff
@@ -1,5 +1,22 @@
...
 // Client::UI::RaptureAtkColorDataManager
+// Used by the ColorPicker addon
 [StructLayout(LayoutKind.Explicit, Size = 0x18)]
-public struct RaptureAtkColorDataManager;
+public unsafe struct RaptureAtkColorDataManager {
+    [FieldOffset(0x08)] public ColorSet* ColorSets;
+    [FieldOffset(0x10)] public int NumColorSets;
+    [FieldOffset(0x14)] public bool IsLoaded;
+
+    public Span<ColorSet> ColorSetsSpan => new(ColorSets, NumColorSets);
```
wolfcomp, line 11: Maybe return an empty span if the loaded is false? Or is `NumColorSets` 0 at that moment?

PR #1307, opened 2025-02-15
Fix offset of AtkArrayData.SubscribedAddonsCount

Haselnussbomber:
The SubscribedAddons array most likely got bigger, but I still can't find a length in the exe.
Anyhow, that's why the offset of `SubscribedAddonsCount` shifted.


Haselnussbomber:
Nevermind, that was me. LMAO

PR #1326, opened 2025-03-24
materialize other various additions

Jaksuhn:


FFXIVClientStructs/FFXIV/Client/Game/Event/EventFramework.cs
```diff
@@ -92,3 +100,9 @@ public enum ContentType : byte {
     Public,
     GoldSaucer
 }
+
+public enum MaterializeEventId {
```
Haselnussbomber, line 104: See https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/Event/EventHandler.cs#L131

Jaksuhn: Is there a reason those weren't broken up in `EventHandlerContent`?

Haselnussbomber: Because EventHandlerContent is a ushort.

Jaksuhn: Well, would you rather the enum not exist at all, or moved to `EventHandler`? I figured in general it's better to have less "pass random value" functions and have backing enums

PR #1327, opened 2025-03-24
Remove `[Obsolete(error: true)]` fields from Texture

Exter-N:


Haselnussbomber:
Maybe earlier, with 7.21. I'm not exactly sure how the plan is.
Usually we let them sit for a couple weeks after a patch, so that devs can make the necessary changes.

wolfcomp:
Just going to close this as it is going to be one commit later one.

PR #1341, opened 2025-03-26
Fix render camera offsets for 7.2

AsgardXIV:
Calculated this based on the movement of NearPlane/FarPlane, which I use, others are an educated guess as ProjectionMatrix is correct.


AsgardXIV:
> Field overlap detected in Client::Game::UI::Map with field LevequestMarkers

I don't think I introduced this?

Haselnussbomber:
No, Cara pushed 79e665e07f25c7b7c91f5107190b71b8147f2c45 without going through the CI check. Ignore it.^^

PR #1345, opened 2025-03-27
Refactor: WeaponState, AutoAttackState

mariamatthews:
- Redefined IsUnsheathed (0x00), SheatheCooldown (0x04), AutoSheathTimer (0x08), and AutoSheatheState (0x0C).
- Deprecated the embedded AutoAttackState at 0x10 (now managed globally via DAT_142921510).
- Added Tick, SetUnsheathed, and SetUnsheathed2 and other methods with updated signatures.
- Revised AutoAttackState interop with SetImpl, Get, and a global binding via Instance().
- Updated data.yml to reflect context and current implementation.


FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
```diff
@@ -1,21 +1,142 @@
...
+    public static ref AutoAttackState Global
+    {
+        get
+        {
+            unsafe
+            {
+                return ref Unsafe.AsRef<AutoAttackState>(Instance());
+            }
+        }
+    }
```
wolfcomp, line 141: Remove this as people can go through `UIState.Instance()->WeaponState.AutoAttackState` to get into here

ida/data.yml
```diff
@@ -2125,10 +2125,9 @@ classes:
       0x140991610: SetUnsheathed2
   Client::Game::UI::AutoAttackState:
     funcs:
-      0x140991770: Get
-      0x1409917C0: OnActorControl
```
wolfcomp, line 2129: This still exists why remove it?

ida/data.yml
```diff
@@ -2125,10 +2125,9 @@ classes:
...
+      0x140991970: Get
```
wolfcomp, line 2128: Wrong offset for this previous is correct.

ida/data.yml
```diff
@@ -2125,10 +2125,9 @@ classes:
...
       0x140991830: SetImpl
-      0x140991970: Set
+      0x140991770: Set
```
wolfcomp, line 2130: Wrong offset for this previous is correct.

wolfcomp (changes requested):
Everything listed has to be changed before even looking at the sigs.

Haselnussbomber:
Heya. Just in case you don't know: the GitHub Action is is comparing your branch with the current main branch.
Since yours is older it's missing things, so it complains about it.
You can rebase or merge main onto your branch, or ignore what it posted.

mariamatthews:
> Heya. Just in case you don't know: the GitHub Action is is comparing your branch with the current main branch. Since yours is older it's missing things, so it complains about it. You can rebase or merge main onto your branch, or ignore what it posted.

Hi @Haselnussbomber! I thought it might be the case - I should have merged main in first, apologies!
Let me know what do you think about the new code update :)

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
mariamatthews: Makes sense :) Done!

ida/data.yml
mariamatthews: Apologies, I think it was very late night evening. Thanks for spotting it!

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
```diff
@@ -1,11 +1,10 @@
-using System.Runtime.CompilerServices;
 namespace FFXIVClientStructs.FFXIV.Client.Game.UI;
 // Client::Game::UI::WeaponState
 /// <summary>
 /// Represents the in-memory layout of the WeaponState structure.
 /// </summary>
 [GenerateInterop]
-[StructLayout(LayoutKind.Explicit, Size = 0x18)]
+[StructLayout(LayoutKind.Explicit, Size = 0x10)]
```
wolfcomp, line 7: The previous size is correct.

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
```diff
@@ -27,13 +26,6 @@ public partial struct WeaponState {
     [FieldOffset(0x0C)]
     public bool AutoSheatheState;
 
-    // Offset 0x10: Originally, this offset was used for AutoAttackState.
-    // In older versions, a simple bool (IsAutoAttacking) resided here.
-    // In the current patch, analysis shows auto-attack state is handled globally via DAT_142921510.
-    [FieldOffset(0x10)]
-    [Obsolete("AutoAttackState is now managed globally via DAT_142921510. Use AutoAttackState.Global instead.", true)]
-    public AutoAttackState AutoAttackState;
-
```
wolfcomp: I never specified to remove the field, only the obsolete.

wolfcomp (changes requested):
Task failed.

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
```diff
@@ -1,21 +1,129 @@
...
+
+    /// <summary>
+    /// Handles actor control updates related to AutoAttack state.
+    /// Note: Internally accesses LocalPlayer at offset 0x9a0 in version 7.0 (previously 0x9b0).
+    /// </summary>
+    /// <param name="statePtr">Pointer to the byte representing the new state.</param>
+    /// <param name="newState">New state value.</param>
+    /// [MemberFunction("88 11 80 FA")] shorter but risky
+    [MemberFunction("88 11 80 FA 01 75 ?? 48 8B 0D ?? ?? ?? ?? 48 85 C9 74 ?? 38 15 ?? ?? ?? ?? 74 ?? 48 81 C1 ?? ?? ?? ??")]
+    public partial void OnActorControl(ref byte statePtr, byte newState);
```
wolfcomp, line 112: This function is defined wrong, it only has 1 arg for definition in C# where
`a2` is a bool of the new state

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
```diff
@@ -1,21 +1,129 @@
...
+
+    /// <summary>
+    /// Sets the auto-attack state after performing configuration and validation checks.
+    /// Wraps the SetImpl call after ensuring conditions are met.
+    /// </summary>
+    [MemberFunction("E8 ?? ?? ?? ?? B0 ?? EB ?? 48 8D 0D")]
+    public partial bool Set(ulong param1, byte param2);
```
wolfcomp, line 119: This function is defined wrong, it only has 1 arg for definition in C# where
`a2` is a bool of the new state

wolfcomp (changes requested):
Some wrong defined methods that needs to be fixed.

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
mariamatthews: Right! Is that because in Ghidra the first parameter (param_1) typically represents the this pointer (the pointer to the struct itself), and therefore in C# we don't explicitly define it?

Haselnussbomber: Yes exactly. Member functions are getting a pointer to the instance passed as first argument (`thisPtr`).

mariamatthews: Right, thanks for clarifying that!

mariamatthews:
Quick question about the automated check reporting:

**Breaking Changes  
Member exists in left but not in right  
WeaponState: bool IsAutoAttacking**

This seems to be triggered because the original metadata defined this offset as a simple bool, whereas we now explicitly define it as a struct (**AutoAttackState**).

Is this something I need to address, or is it safe to ignore?

Haselnussbomber:
It's fine in this case and safe to ignore.
This field had the Obsolete attribute with error set to true on it. Due to the API bump plugins have to be recompiled and that error would prevent a successful compilation, until the code was changed to not use that field anymore.
We would have removed it soon anyway.

FFXIVClientStructs/FFXIV/Client/Game/UI/WeaponState.cs
```diff
@@ -1,21 +1,129 @@
...
+/// </summary>
+[GenerateInterop]
 [StructLayout(LayoutKind.Explicit, Size = 0x18)]
-public struct WeaponState {
-    [FieldOffset(0x00)] public bool IsUnsheathed;
-    [FieldOffset(0x04)] public float SheatheCooldown;
-    [FieldOffset(0x08)] public float AutoSheathTimer;
-    [FieldOffset(0x0C)] public bool AutoSheatheState;
-    [FieldOffset(0x10)] public AutoAttackState AutoAttackState; // note: not sure whether this is actually a part of this structure or next field of a parent
-    [FieldOffset(0x10), Obsolete("Use AutoAttackState.IsAutoAttacking", true)] public bool IsAutoAttacking;
```
wolfcomp, line 11: Do not remove the obsoleted field here. These are slated for removal later.

PR #1358, opened 2025-03-28
Enforce x64 builds

Haselnussbomber:
It doesn't make sense to allow x86 builds of CS, so we should limit it to x64.

This also aligns it with the [Dalamud.NET.Sdk](https://github.com/goatcorp/Dalamud.NET.Sdk/blob/master/Dalamud.NET.Sdk/Sdk/Sdk.props#L38) to avoid mixing Any CPU and x64 configurations in Solution files.


Haselnussbomber:
This actually makes me specify `<Platform Project="x64" />` on every project in the slnx. I don't know.
If you guys still think this is a good idea, just reopen the PR.

PR #1366, opened 2025-03-30
Update Loot.cs

Taurenkey:
Adds WeeklyLootItem to LootItem

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/UI/Loot.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/UI/Loot.cs
@@ -31,6 +31,7 @@ public struct LootItem {
     [FieldOffset(0x20)] public RollState RollState;
     [FieldOffset(0x24)] public RollResult RollResult;
     [FieldOffset(0x28)] public byte RollValue;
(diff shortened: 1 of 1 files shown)


FFXIVClientStructs/FFXIV/Client/Game/UI/Loot.cs
```diff
@@ -31,6 +31,7 @@ public struct LootItem {
     [FieldOffset(0x20)] public RollState RollState;
     [FieldOffset(0x24)] public RollResult RollResult;
     [FieldOffset(0x28)] public byte RollValue;
+    [FieldOffset(0x34)] public byte WeeklyLootItem; //0 for items that aren't weekly restricted, 1 for those that are.
```
Caraxi, line 34: From the comment it seems like this is a bool?

Taurenkey: Possibly, in all my testing it has either been 0 or 1, but that's not to say I've overlooked something.

Caraxi: The only reference I can find to it, it is used as a bool so likely safe to define it as such

PR #1369, opened 2025-03-30
More Arrays

Glyceri:
Mapped quite a lot more arrays c:


Haselnussbomber (changes requested):
You only need t o put CExportIgnore on the `_data` arrays, because they overlap with all other fields.

PR #1375, opened 2025-04-04
Make ID structs equatable and comparable.

Ottermandias:
See title. No clue if this is breaking. No properties are removed in any case, so as long as adding methods and interfaces to structs is fine, this should be no issue.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/HousingManager.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/HousingManager.cs
@@ -147,7 +147,7 @@ public unsafe partial struct HousingManager {
 }
 
 [StructLayout(LayoutKind.Explicit, Size = 0x8)]
-public struct HouseId {
+public struct HouseId : IEquatable<HouseId>, IComparable<HouseId> {
     /// <remarks>
     /// Masked data:<br/>
     /// - <c>0b1000_0000</c> (<c>0x80</c>) = Apartment Flag<br/>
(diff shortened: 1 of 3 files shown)


FFXIVClientStructs/FFXIV/Client/Game/HousingManager.cs
```diff
@@ -165,6 +165,8 @@ public struct HouseId {
     [FieldOffset(0x4)] public ushort TerritoryTypeId;
     [FieldOffset(0x6)] public ushort WorldId;
 
+    [FieldOffset(0x0), CExportIgnore] public long Id;
```
wolfcomp, line 168:
```suggestion
    [FieldOffset(0x0), CExportIgnore] private long Id;
```

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -155,12 +155,20 @@ public unsafe partial struct GameObject {
 //   if (BaseId != 0) ObjectId = BaseId, Type = 1
 // else ObjectId = EntityId, Type = 0
 [StructLayout(LayoutKind.Explicit, Size = 0x8)]
-public struct GameObjectId {
+public struct GameObjectId : IEquatable<GameObjectId>, IComparable<GameObjectId> {
     [FieldOffset(0x0)] public uint ObjectId;
     [FieldOffset(0x4)] public byte Type;
 
-    public static unsafe implicit operator ulong(GameObjectId id) => *(ulong*)&id;
+    [FieldOffset(0x0), CExportIgnore] public ulong Id;
```
wolfcomp, line 162:
```suggestion
    [FieldOffset(0x0), CExportIgnore] private ulong Id;
```

wolfcomp (commented):
Should probably have all the type fields be private that is the same size as the struct, so people don't try and use them without understanding it is a combined field type.

EventId will get the same treatment as these two suggestions have in 7.3 as i didn't think about this before now.

Haselnussbomber:
I think it's fine to have these Id fields. We have Value fields in various structs already (see `CStringPointer`, `Pointer<T>`, `WeaponModelId`, `EquipmentModelId`...).
The only change I would like to see here is that fields should be ordered by offset.

PR #1409, opened 2025-04-27
InteropGenerator: support multiple inheritance pointer arithmetic

salanth357:
Adds a static `T* FromBasePointer(TB* basePtr)` that uses a known static vtbl pointer to validate that basePtr is pointing to a TB baseclass inside of a T.

The use on AtkComponentTextInput is meant to be the first usage of it, but right now the interface type doesn't have any of its vfuncs reversed- I can work on adding some if needed, but for now, this is sufficient to enable the autocomplete functionality in dalamud


aers:
It's a good start. I'm thinking about trying something more elaborate with RTTI in general, will try to get back to you in a few days.

PR #1412, opened 2025-04-29
Update data.yml

SakuraFutaba:
functions related to `ProcessPacket`

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -335,8 +335,16 @@ functions:
   0x1400BA490: IsPartyMemberByNameOrContentId # (name, contentId)
   0x1400BA5C0: GetPartyMemberClassJobByContentId
   0x140844B40: ObjectIdToEntityId # returns invalid id if objectid is not a player
+  0x140A210E0: SendEventCompletePacket
+  0x140A21C20: ProcessPacketEventStart
+  0x140A21C80: ProcessPacketEventFinish
+  0x140A21D40: ProcessPacketBattleTalk
+  0x140A21DE0: ProcessPacketBalloonTalk
   0x140A21A00: ProcessPacketEventPlay
   0x140A21EE0: ProcessPacketSystemLogMessage
+  0x140A21FA0: ProcessPacketMapUpdate
+  0x140A22060: ProcessPacketFateInfo
+  0x140A220B0: ProcessPacketFateProgress
   0x140A251B0: SendClientSelectDataConfigFlagsUpdate
   0x140A25AE0: ProcessPacketSpawnPlayer
   0x140A25BE0: ProcessPacketSpawnNPC
@@ -396,6 +404,9 @@ functions:
   0x140A396B0: ProcessPacketUpdatePartyPositions
   0x140A396E0: ProcessPacketUpdateAllianceNormalPositions
   0x140A39710: ProcessPacketUpdateAllianceSmallPositions
+  0x140A39830: ProcessPacketChatReceived
+  0x140A399E0: ProcessPacketInventoryTransaction
+  0x140A39A00: ProcessPacketItemInfo
   0x140A39B90: ProcessPacketRSVData
   0x140A39BB0: ProcessPacketRSFData
   0x140A39C10: ProcessPacketHuntingLogEntry


SakuraFutaba:
Additionally, I found that the function `ProcessPacketNpcYell` calls `g_Client::Game::UI::GoldSaucerYell_Instance`. Maybe `GoldSaucerYell` should be renamed to `NpcYell`.

ida/data.yml
```diff
@@ -396,10 +404,25 @@ functions:
...
+  0x140A399E0: ProcessPacketInventoryTransaction
+  0x140A39A00: ProcessPacketItemInfo
   0x140A39B90: ProcessPacketRSVData
   0x140A39BB0: ProcessPacketRSFData
   0x140A39C10: ProcessPacketHuntingLogEntry
   0x140A39C50: ProcessPacketNPCYell
+  0x140A39D10: ProcessPacketEquipDisplayFlags
+  0x140A39DE0: ProcessPacketCompletedAchievements
+  0x140A39E30: ProcessPacketLandSetInitialize
+  0x140A3B4D0: ProcessPacketDailyQuests
```
Haselnussbomber, line 417: This address resolves to something GcArmyExpedition (Squadron) related.
Couldn't find anything matching its name up to 10 versions back.

ida/data.yml
```diff
@@ -396,10 +404,25 @@ functions:
...
+  0x140A39830: ProcessPacketChatIncoming
+  0x140A399E0: ProcessPacketInventoryTransaction
+  0x140A39A00: ProcessPacketItemInfo
   0x140A39B90: ProcessPacketRSVData
   0x140A39BB0: ProcessPacketRSFData
   0x140A39C10: ProcessPacketHuntingLogEntry
   0x140A39C50: ProcessPacketNPCYell
+  0x140A39D10: ProcessPacketEquipDisplayFlags
+  0x140A39DE0: ProcessPacketCompletedAchievements
+  0x140A39E30: ProcessPacketLandSetInitialize
```
Haselnussbomber, line 416: This address resolves to something Achievement related.
It seems to be matching an address in 7.2hf1. Here the correct address for 7.21:
```suggestion
  0x140A3A510: ProcessPacketLandSetInitialize
```

Haselnussbomber (changes requested):
Can you have a look at those 2 addresses again?

ida/data.yml
SakuraFutaba: 0x140A3B4D0: ProcessPacketGcArmyExpeditionMissionResult
I have renamed it. This function calls when Squadron missions finished and interact with squadron sergeant.

ida/data.yml
```diff
@@ -1041,11 +1041,20 @@ classes:
       0x140A3DBF0: ctor
       0x140A3E6A0: Finalizer
       0x140A3E740: Initialize
+      0x140A3ED50: ProcessPacketInventoryTransaction
+      0x140A3F110: ProcessPacketMarketBoardInfo
```
Haselnussbomber, line 1045: This updates [InventoryManager.RetainerMarketPrices](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs#L27).

ida/data.yml
```diff
@@ -396,10 +404,31 @@ functions:
...
+  0x140A3B4D0: ProcessPacketGcArmyExpeditionMissionResult
+  0x140A3B510: ProcessPacketGcArmyExpeditionEnlistment
+  0x140A3B610: ProcessPacketLandSetMap
+  0x140A3B650: ProcessPacketMissionStart
+  0x140A3B710: ProcessPacketGcArmyExpeditionMissionRepeatFlags
+  0x140A3B9A0: ProcessPacketPlayMotionSync
+  0x140A3BDF0: ProcessPacketMount
+  0x140A3C5A0: ProcessPacketPrepareZoning
+  0x140A3C600: ProcessPacketFall
+  0x140A3C660: ProcessPacketMSQTrackerComplete
```
Haselnussbomber, line 430: The "MSQTracker" is called ScenarioTree. At the very end, it also seems to save some value from the ScenarioTree sheet (column Unknown0) into an MsqProgress config option (under certain conditions). I didn't even know that's a thing. 🤔

Haselnussbomber (changes requested):
I have some more comments after looking at those functions, but I'd like to make clear, that:
1\) I'm not a maintainer - just someone who contributes a lot and looks at things.
2\) I didn't test any of these addresses, so I can only guess as to what they do based on static analysis.

ida/data.yml
SakuraFutaba: Thank you for your review!
By the way, I noticed that some of these functions eventually call
`Component::Exd::ExdModule_GetRowBySheetIndexAndRowId(*(_QWORD *)(g_Client::System::Framework::Framework_InstancePointer2 + 11072), 612i64, a1);`
Is there a way to find out what the sheet name is for a given `SheetIndex`? (In this case 612)

Haselnussbomber: Yes, sheet indexes are in a csv file in the sqpack file `0a0000` inside the `exd/root.exl` file.
612 is ScenarioTree

SakuraFutaba: Thanks a lot! I can now successfully read the `exd/root.exl` file with FFXIV Explorer.

PR #1418, opened 2025-05-02
Added function that attaches pointers to the model

Glorou:
Just added some notes and a sig for the function attaches the most of the pointers to the model when first rendered in, and initializes some fields e.g. shapekey and attribute masks

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Render/ModelRenderer.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Render/ModelRenderer.cs
@@ -103,4 +103,13 @@ public unsafe partial struct ModelRenderer {
     // This function, among other things, constructs an OnRenderMaterialParams struct with its params and calls CharacterBase.OnRenderMaterial with it (through some indirections - see Model.RenderMaterialCallback).
     [MemberFunction("E8 ?? ?? ?? ?? 44 0F B7 28")]
     public partial ushort* OnRenderMaterial(ushort* outFlags, OnRenderModelParams* param, Material* material, uint materialIndex);
+    
+    // This is called by Client::Graphics::Render::Model.SetupFromModelResourceHandle to set up the models pointers, before drawing, local resources (from the stack) are
+    // res8 = Render.Skeleton 
+    // res10 = Model Resource Handle 
+    // res18 = Render.Skeleton (??? exact same as res8, in case it needs to regenerate for a different skeleton?)  
+    // res20 = Render.Model
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 75 ?? 48 8B 07 48 8B CF FF 50 ?? 32 C0")]
+    public partial uint ModelDrawInit(Model* data, ModelResourceHandle* mdlHandle, Callback _renderModelCallback, Callback *_renderMaterialCallback);
+
 }


FFXIVClientStructs/FFXIV/Client/Graphics/Render/ModelRenderer.cs
```diff
@@ -103,4 +103,13 @@ public enum WellKnownSampler {
...
     [MemberFunction("E8 ?? ?? ?? ?? 44 0F B7 28")]
     public partial ushort* OnRenderMaterial(ushort* outFlags, OnRenderModelParams* param, Material* material, uint materialIndex);
+    
+    // This is called by Client::Graphics::Render::Model.SetupFromModelResourceHandle to set up the models pointers, before drawing, local resources (from the stack) are
+    // res8 = Render.Skeleton 
+    // res10 = Model Resource Handle 
+    // res18 = Render.Skeleton (??? exact same as res8, in case it needs to regenerate for a different skeleton?)  
+    // res20 = Render.Model
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 75 ?? 48 8B 07 48 8B CF FF 50 ?? 32 C0")]
+    public partial uint ModelDrawInit(Model* data, ModelResourceHandle* mdlHandle, Callback _renderModelCallback, Callback *_renderMaterialCallback);
```
Haselnussbomber, line 113: This is a member function of `Client::Graphics::Render::Model`, not from `ModelRenderer`, so this shouldn't be here.

Besides that:
- The first argument should be omitted, since it's not a static function (the generator will add `Model* thisPtr` automatically).
- The third argument should be a pointer (`Callback*`).
- The return type should be bool, as it only returns 0 or 1, which is checked with `if ( !sub_140404090(a1, a2, a4, a5) )`.

Please add this to [Model](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Graphics/Render/Model.cs) instead:
```cs
    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 75 ?? 48 8B 07 48 8B CF FF 50 ?? 32 C0")]
    public partial bool ModelDrawInit(ModelResourceHandle* mdlHandle, Callback* renderModelCallback, Callback* renderMaterialCallback);
```

Haselnussbomber (changes requested):
Yo, I have something to say about this. 🤓

FFXIVClientStructs/FFXIV/Client/Graphics/Render/Model.cs
```diff
@@ -32,4 +32,8 @@ public unsafe partial struct Model {
 
     public ReadOnlySpan<Pointer<Material>> MaterialsSpan
         => new(Materials, MaterialCount);
+    
+    
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 75 ?? 48 8B 07 48 8B CF FF 50 ?? 32 C0")]
+    public partial bool ModelDrawInit(ModelResourceHandle* mdlHandle, ModelRenderer.Callback* renderModelCallback, ModelRenderer.Callback* renderMaterialCallback);
```
wolfcomp, line 38: Add the related function to the correct location in `data.yml`

wolfcomp (changes requested):

Glorou:
Sorry forgot about this change request, should be fixed now I think

PR #1420, opened 2025-05-03
Add OpenListingByContentId to AgentLookingForGroup

AtmoOmen:


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentLookingForGroup.cs
```diff
@@ -45,6 +45,9 @@ public unsafe partial struct AgentLookingForGroup {
     [MemberFunction("48 89 5C 24 ?? 57 48 83 EC ?? 48 8B FA 48 8B D9 E8 ?? ?? ?? ?? 48 8B 8B ?? ?? ?? ?? 48 85 C9")]
     public partial bool OpenListing(ulong listingId);
 
+    [MemberFunction("40 53 48 83 EC 20 48 8B D9 E8 ?? ?? ?? ?? 84 C0 74 07 C6 83 90 31 00 00 01")]
+    public partial bool OpenListingByContentId(ulong contentId);
```
wolfcomp, line 49: Add this function to the correct location in `data.yml`

wolfcomp (changes requested):

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentLookingForGroup.cs
wolfcomp, line 49:
```suggestion
    public partial bool OpenListingByContentId(ulong contentId); // Actual call inlined
```

wolfcomp (changes requested):
Just one more change so we can traceback if it ever gets lost during a patch.

PR #1427, opened 2025-05-09
Update data.yml and NetworkModule

SakuraFutaba:
I renamed `ZoneConnection` and `ChatConnection` to `ZoneClient` and `ChatClient` to better match the RTTI naming conventions. 
Open to suggestions if there is a better name.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Application/Network/ChatClient.cs
+++ b/FFXIVClientStructs/FFXIV/Application/Network/ChatClient.cs
@@ -1,0 +1,6 @@
+namespace FFXIVClientStructs.FFXIV.Application.Network;
+
+[GenerateInterop]
+[StructLayout(LayoutKind.Explicit, Size = 0xA0)]
+public unsafe partial struct ChatClient {
+}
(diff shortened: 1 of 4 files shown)


FFXIVClientStructs/FFXIV/Application/Network/ChatClient.cs
```diff
@@ -0,0 +1,6 @@
+namespace FFXIVClientStructs.FFXIV.Application.Network;
+
+[GenerateInterop]
+[StructLayout(LayoutKind.Explicit, Size = 0xA0)]
```
pohky, line 4: the size of this is 0xA8

FFXIVClientStructs/FFXIV/Application/Network/ZoneClient.cs
```diff
@@ -0,0 +1,6 @@
+namespace FFXIVClientStructs.FFXIV.Application.Network;
+
+[GenerateInterop]
+[StructLayout(LayoutKind.Explicit, Size = 0x98)]
```
pohky, line 4: and the size of this is 0xA0

pohky (changes requested):

PR #1436, opened 2025-05-27
add Occult Crescent ContentValues

Cytraen:


Haselnussbomber:
[Inspect](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/UI/Inspect.cs#L61-L69) probably has it too?

PR #1449, opened 2025-06-14
Fix AtkExternalInterface.CallHandler return value

Haselnussbomber:
Without the return value, hooking `AtkExternalInterface.CallHandler` will cause a crash.

I'm not sure if `AtkExternalInterface.PlaySoundEffect` and `AtkModuleEvent.CallHandler` are also missing the return value, so I didn't include them.


wolfcomp:
`AtkModuleEvent.CallHandler` is 100% missing the return value.
image

`AtkExternalInterface.PlaySoundEffect` is supposed to be void.
image
```
mov     eax, [rsp+38h+var_18.Type]
shr     eax, 5
test    al, 1
``` is just this part
![image](https://github.com/user-attachments/assets/9c109c86-8e5a-40e0-94bb-be778a4fdb1a)
and isn't used as the return part due to the register not being a "valid" return type that is expected from the assembly, unless there is some MSVC optimizations I haven't seen yet.

PR #1458, opened 2025-06-19
Update AgentLoot offsets, Add SelectedReward to AgentContentsFinder and some new agents

Nukoooo:


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInspect.cs
```diff
@@ -21,7 +21,13 @@ public unsafe partial struct AgentInspect {
     [FieldOffset(0x108)] public Utf8String OnlineId;
     [FieldOffset(0x170), FixedSizeArray] internal FixedSizeArray3<Utf8String> _chocoboBarding;
     [FieldOffset(0x2A8), FixedSizeArray] internal FixedSizeArray13<ItemData> _items;
-    [FieldOffset(0x448)] public FreeCompanyData FreeCompany;
+
+    [FieldOffset(0x449)] public bool IsPartOfFreeCompany;
```
Haselnussbomber: From what I can see, this should be named `IsBuddyInspect` or something along the lines.
It is set via the `ExamineCharacter` function and is checked in the `Show` to decide whether to open the addon `BuddyInspect` or `CharacterInspect`.

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInspect.cs
```diff
@@ -48,15 +54,16 @@ public unsafe partial struct AgentInspect {
     [StructLayout(LayoutKind.Explicit, Size = 0x318)]
     public partial struct InspectCharaView;
 
-    [StructLayout(LayoutKind.Explicit, Size = 0x88)]
+    [StructLayout(LayoutKind.Explicit, Size = 0x80)]
     public struct FreeCompanyData {
         //[FieldOffset(0x00)] public byte Unkown4b0; // Maybe FreeCompany get status 1 = Finished
+        [Obsolete("Use IsPartOfFreeCompany instead")]
```
Haselnussbomber: Since this is no longer part of this struct, the returned data is no longer valid. Please set the error flag to true.
Also, besides from being a confusing obsolete message, the game checks if the Id is not 0 to detect if the player is in a Free Company (~~at 0x140E78654, and 0x140E7AC6E~~).

Haselnussbomber (changes requested):

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInspect.cs
Haselnussbomber: Dammit, I had an old IDA db open. Offsets are 0x140E788F4 and 0x140E7AF0E.

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInspect.cs
Nukoooo: If that is the case, then 0x448 could be ``IsCharacterInspect``?

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInspect.cs
Nukoooo: Hope this one is clear, I suck at this

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentInspect.cs
Haselnussbomber: Nope. Both BuddyInspect and CharacterInspect send a callback at the very end of their OnSetup function with Int = 0, which makes the agent set this field to true (at 0x140E788BF). It's used in the agents Update function to check if the addon is set up before calling RefreshAddon.

PR #1466, opened 2025-06-28
Add GetClassJobLevel to PlayerState

AtmoOmen:
for convenience

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
@@ -280,7 +280,7 @@ public unsafe partial struct PlayerState {
 
     /// <summary> Get the level of a specific job. </summary>
     [MemberFunction("E8 ?? ?? ?? ?? 0F B6 0D ?? ?? ?? ?? 4C 8D 3D")]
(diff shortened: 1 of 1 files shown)


FFXIVClientStructs/FFXIV/Client/Game/UI/PlayerState.cs
```diff
@@ -278,6 +278,10 @@ public uint GetContentValue(uint key) {
     /// <returns> The 0-based value of the pose. </returns>
     public byte CurrentPose(EmoteController.PoseType pose) => !Enum.IsDefined(pose) ? (byte)0 : SelectedPoses[(int)pose];
 
+    /// <summary> Get the level of a specific job. </summary>
+    [MemberFunction("E8 ?? ?? ?? ?? 0F B6 0D ?? ?? ?? ?? 4C 8D 3D")]
+    public partial short GetClassJobLevel(uint classJobId, bool checkParentJob);
```
wolfcomp:
```suggestion
    public partial short GetClassJobLevel(uint classJobId, bool shouldGetSynced);
```

wolfcomp (changes requested):
Just correct one thing to make it more clear what it is actually returning if the `bool` is true

PR #1469, opened 2025-06-30
Some additions

Nukoooo:


FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
```diff
@@ -56,6 +56,12 @@ public unsafe partial struct InventoryManager {
     [MemberFunction("E8 ?? ?? ?? ?? EB 7A 83 F8 04")]
     public partial int MoveItemSlot(InventoryType srcContainer, ushort srcSlot, InventoryType dstContainer, ushort dstSlot, byte unk = 0);
 
+    [MemberFunction("40 55 53 56 57 41 55 41 57 48 8D 6C 24 ?? 48 81 EC ?? ?? ?? ?? 48 8B 05 ?? ?? ?? ?? 48 33 C4 48 89 45 ?? 8D B2")]
+    public partial int SplitItem(InventoryType container, ushort slot, int quantity);
+
+    [MemberFunction("40 56 57 41 57 48 83 EC ?? 45 0F BF F8")]
+    public partial int DiscardItem(InventoryType container, ushort slot);
+
```
wolfcomp:
```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? 48 8D 8B ?? ?? ?? ?? 48 8B 11")]
    public partial int DiscardItem(InventoryType container, ushort slot);

```

Use jump sig instead of function start if available as these can help trace back the function if it breaks.

wolfcomp (changes requested):
Also add the two functions to `data.yml`

Nukoooo:
Unfortunately I only have the exe for 7.25 not 7.25hf2 so I can't really add these two functions to data.yml

PR #1471, opened 2025-07-01
AtkComponentNumericInput

MidoriKami:
Adds the inner set value function that allows us to set a value without triggering the sfx

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentNumericInput.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentNumericInput.cs
@@ -18,4 +18,9 @@ public unsafe partial struct AtkComponentNumericInput : ICreatable {
 
     [MemberFunction("E8 ?? ?? ?? ?? 81 4F ?? ?? ?? ?? ?? EB 07")]
     public partial void SetValue(int value);
+
+    [MemberFunction("E9 ?? ?? ?? ?? 33 D2 F7 F1")]
+    public partial void SetValue(int value, bool triggerCallback, bool playSoundEffect);
+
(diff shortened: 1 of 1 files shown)


MidoriKami:
The generator doesn't like overloads? 🤔

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentNumericInput.cs
```diff
@@ -18,4 +18,9 @@ public unsafe partial struct AtkComponentNumericInput : ICreatable {
 
     [MemberFunction("E8 ?? ?? ?? ?? 81 4F ?? ?? ?? ?? ?? EB 07")]
     public partial void SetValue(int value);
+
+    [MemberFunction("E9 ?? ?? ?? ?? 33 D2 F7 F1")]
+    public partial void SetValueInner(int value, bool triggerCallback, bool playSoundEffect);
+
+    public void SetValueSilent(int value) => SetValueInner(value, true, false);
```
wolfcomp, line 25:
```suggestion
    public partial void SetValueInner(int value, bool triggerCallback = true, bool playSoundEffect = false);
```

Also add the function address to `data.yml`

wolfcomp (changes requested):

FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentNumericInput.cs
MidoriKami: I disagree with the proposed change, I considered having the defaults set like that but that wouldn't reflect a typical use of the function.

If you don't like the helper function then I can remove that, but I won't set default args on the function.

Will add to data.yml though

PR #1486, opened 2025-07-13
Add remarks about how quest IDs are truncated

redstrate:
After some exploring, QuestManager truncates uint quest IDs for a reason to transform them into ushorts as seen in these work definitions. I added it as a remark to the ushort quest IDs I found so other people aren't confused why quest IDs don't match up with what you see in the sheet.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Application/Network/WorkDefinitions/DailyQuestWork.cs
+++ b/FFXIVClientStructs/FFXIV/Application/Network/WorkDefinitions/DailyQuestWork.cs
@@ -6,6 +6,7 @@ namespace FFXIVClientStructs.FFXIV.Application.Network.WorkDefinitions;
 [Inherits<Base>]
 [StructLayout(LayoutKind.Explicit, Size = 0x10)]
 public partial struct DailyQuestWork {
+    /// <remarks> Not a row ID in the Quest Sheet. This is only the first two bytes (e.g. row 67540 would be stored as 2004.) </remarks>
     [FieldOffset(0x08)] public ushort QuestId;
     [FieldOffset(0x0A)] public byte Flags;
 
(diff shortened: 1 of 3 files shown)


Haselnussbomber:
The quest ids starting from 0 as ushort are actually the correct/normalized ids. These are transmitted via packets over the network.
The RowId in the Quest sheet is offset by 65536, because then it can reference unlock links and Quest rows at the same time, using one column. It passes the id to [UIState. IsUnlockLinkUnlockedOrQuestCompleted](https://github.com/aers/FFXIVClientStructs/blob/62ea2008fc5a1a88733653925dcea946d512eb1c/FFXIVClientStructs/FFXIV/Client/Game/UI/UIState.cs#L132-L144), which checks the UnlockLinkBitmask if the value is below 65536, otherwise it casts it to ushort to get rid of the upper bits (which would just be the marker that it’s a quest: 0x10000) to make it a normalized quest id for [QuestManager.IsQuestComplete](https://github.com/aers/FFXIVClientStructs/blob/62ea2008fc5a1a88733653925dcea946d512eb1c/FFXIVClientStructs/FFXIV/Client/Game/QuestManager.cs#L87-L88).
In my opinion, the comment explaining that should be on the Quest sheet instead.

redstrate:
> Yeah, I'm also not sure how to do that, but I also doubt anyone would read it anyway - even if it was written somewhere.

This is where I disagree :D I think if we expose it properly, people _will_ read it - see https://github.com/WorkingRobot/EXDViewer/pull/2 where I try to show the few, but useful comments we have in EXDSchema. I opened https://github.com/xivdev/EXDSchema/issues/94 to see if we can extend EXDSchema to support this, and then figure it out from there.

PR #1506, opened 2025-08-06
Fix some addon config offsets

nebel:


Haselnussbomber:
This is already included in https://github.com/aers/FFXIVClientStructs/pull/1504

PR #1527, opened 2025-08-10
Remove PhysicsGroup enum

redstrate:
Lets go ahead and update the obsoletes for 7.3. I also removed the enum describing the PhysicsGroup because:
1. I don't really know if it can be truly exhaustive, or if they reuse the same rows for different purposes.
2. You should really be looking at the Excel sheet instead, because it's literally an index into that.

Whoever was using the enum can copy that into their project instead and deal with figuring out which rows are which.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
@@ -6,7 +6,7 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Physics;
 // Client::Graphics::Physics::BoneSimulator
 [StructLayout(LayoutKind.Explicit, Size = 0x450)]
 public unsafe struct BoneSimulator {
-    [FieldOffset(0x10)] public PhysicsGroup Group;
+    [FieldOffset(0x10)] public uint Group; // Row ID in the PhysicsGroup sheet.
     [FieldOffset(0x18)] public Skeleton* Skeleton; // Client::Graphics::Render::Skeleton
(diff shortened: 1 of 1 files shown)


FFXIVClientStructs/FFXIV/Client/Graphics/Physics/BoneSimulator.cs
```diff
@@ -15,25 +15,11 @@ public unsafe struct BoneSimulator {
     // They are hardcoded to roughly 1/60 (0.016666668) and 60 (59.999996) in cases where the sheet isn't used.
     [FieldOffset(0x50)] public float SimulationTime;
     [FieldOffset(0x54)] public float SimulationTimeInv; // 1/SimulationTime
-    [Obsolete("Use SimulationTimeInv instead", true)]
-    [FieldOffset(0x54)] public float Spring; // Default is ~60, intense jitter happens above that value. Lesser values remove the spring in the bone.
     [FieldOffset(0xF6)] public bool IsStarted; // Flag that is set to true when the simulator starts, and is quickly reset
```
wolfcomp:
```suggestion
    [Obsolete("Use SimulationTimeInv instead", true)]
    [FieldOffset(0x54)] public float Spring; // Default is ~60, intense jitter happens above that value. Lesser values remove the spring in the bone.
    [FieldOffset(0xF6)] public bool IsStarted; // Flag that is set to true when the simulator starts, and is quickly reset
```

Do not remove this as it is slated for removal in 7.31

redstrate: Done

PR #1539, opened 2025-08-15
Add nameplate-related GameObject functions

Exter-N:
I'm not 100% sure this is the best name for these functions, but they look to me like what I wrote on the tin.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
@@ -2,6 +2,7 @@ using FFXIVClientStructs.FFXIV.Client.Game.Event;
 using FFXIVClientStructs.FFXIV.Client.Graphics;
 using FFXIVClientStructs.FFXIV.Client.Graphics.Scene;
 using FFXIVClientStructs.FFXIV.Client.LayoutEngine.Group;
+using FFXIVClientStructs.FFXIV.Client.UI.Arrays;
 using FFXIVClientStructs.FFXIV.Common.Math;
 using EventHandler = FFXIVClientStructs.FFXIV.Client.Game.Event.EventHandler;
 
@@ -216,7 +217,9 @@ public unsafe partial struct GameObject {
     [StructLayout(LayoutKind.Explicit, Size = 8)]
     public struct NamePlateColors {
         [FieldOffset(0), CExportIgnore] public ulong Data;
+        /// <seealso cref="Hud2NumberArray.TargetBarBackdropColor"/>
         [FieldOffset(0)] public ByteColor EdgeColor;
+        /// <seealso cref="Hud2NumberArray.TargetBarFillColor"/>
         [FieldOffset(4)] public ByteColor Color;
         
         public static implicit operator ulong(NamePlateColors colors) => colors.Data;


FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -146,6 +146,23 @@ public unsafe partial struct GameObject {
 
     [MemberFunction("E8 ?? ?? ?? ?? 89 85 ?? ?? ?? ?? 0F 57 C0")]
     public partial uint GetObjStrId();
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 85 FF 0F 84 ?? ?? ?? ?? F3 0F 10 97")]
+    public partial float* GetNameplateWorldPosition(float* vector);
```
wolfcomp:
```suggestion
    public partial Vector3* GetNameplateWorldPosition(Vector3* vector);
```

Make sure the `Vector3` is from `Common.Math` namespace in CS

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -146,6 +146,23 @@ public unsafe partial struct GameObject {
...
+
+    public Vector3 GetNameplateWorldPosition() {
+        var vector = stackalloc float[3];
+        GetNameplateWorldPosition(vector);
+        return new Vector3(vector[0], vector[1], vector[2]);
+    }
```
wolfcomp: This should be moved to the individual user to handle

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -146,6 +146,23 @@ public unsafe partial struct GameObject {
...
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 89 43 FB")]
+    public partial ulong GetTargetColorsRaw();
+
+    public (uint BackdropColor, uint FillColor) GetTargetColors() {
+        var colors = GetTargetColorsRaw();
+        return (unchecked((uint)colors), unchecked((uint)(colors >> 32)));
+    }
```
wolfcomp: Is this returned as `ARGB` like the rest of the game with the colors?

Exter-N: Given it can be passed straight to ImGui in plugins, it's `R | G << 8 | B << 16 | A << 24`.

No idea if it's "like the rest of the game".

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
Exter-N: Would it really be correct, as CS' `Vector3` is 16 bytes, and the function takes and returns a 12-byte `float` array?

wolfcomp: The stack alloc in the function above is 16 bytes large, but you are only using the 12 bytes due to it being padding for the last 4 of the Vector3

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
Exter-N: I removed that (even though I think it would be more practical for callers to have it)

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -203,6 +203,20 @@ public unsafe partial struct GameObject {
...
+    public (ByteColor BackdropColor, ByteColor FillColor) GetNamePlateColors() {
+        var colors = GetNamePlateColorsRaw();
+        return (new ByteColor { RGBA = unchecked((uint)colors) }, new ByteColor { RGBA = unchecked((uint)(colors >> 32)) });
+    }
+
+    /// <summary>Gets the world position where the base (bottom center) of the nameplate is to be placed.</summary>
+    /// <param name="vector">A caller-supplied buffer (of one vector) to write the position into.</param>
+    /// <returns><paramref name="vector" /></returns>
+    [MemberFunction("E8 ?? ?? ?? ?? 48 85 FF 0F 84 ?? ?? ?? ?? F3 0F 10 97")]
+    public partial Vector3* GetNamePlateWorldPosition(Vector3* vector);
```
wolfcomp:
```suggestion

    [MemberFunction("E8 ?? ?? ?? ?? 48 89 43 FB")]
    public partial NamePlateColors GetNamePlateColorsRaw();

    /// <summary>Gets the world position where the base (bottom center) of the nameplate is to be placed.</summary>
    /// <param name="vector">A caller-supplied buffer (of one vector) to write the position into.</param>
    /// <returns><paramref name="vector" /></returns>
    [MemberFunction("E8 ?? ?? ?? ?? 48 85 FF 0F 84 ?? ?? ?? ?? F3 0F 10 97")]
    public partial Vector3* GetNamePlateWorldPosition(Vector3* vector);

    [StructLayout(LayoutKind.Explicit), Size = 8]
    public struct NamePlateColors {
        [FieldOffset(0), CExporterIgnore] public ulong Data;
        [FieldOffset(0)] public ByteColor BackdropColor;
        [FieldOffset(4)] public ByteColor FillColor;
        
        public static implicit operator ulong(NamePlateColors colors) => colors.Data;
        public static implicit operator NamePlateColors(ulong colors) => *(NamePlateColors)&colors;
    }
```

wolfcomp (changes requested):
Instead of having the helper function

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -203,6 +203,25 @@ public unsafe partial struct GameObject {
     /// </remarks>
     [MemberFunction("48 89 5C 24 ?? 48 89 6C 24 ?? 48 89 74 24 ?? 57 48 83 EC 20 48 8B 35 ?? ?? ?? ?? 48 8B F9")]
     public partial byte GetNamePlateColorType();
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 89 43 FB")]
+    public partial NamePlateColors GetNamePlateColorsRaw();
```
wolfcomp:
```suggestion
    public partial NamePlateColors GetNamePlateColors();
```

Should match the name in `data.yml` as it is faster to lookup when setting the fields from the structs import

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -203,6 +203,25 @@ public unsafe partial struct GameObject {
...
+    /// <param name="vector">A caller-supplied buffer (of one vector) to write the position into.</param>
+    /// <returns><paramref name="vector" /></returns>
+    [MemberFunction("E8 ?? ?? ?? ?? 48 85 FF 0F 84 ?? ?? ?? ?? F3 0F 10 97")]
+    public partial Vector3* GetNamePlateWorldPosition(Vector3* vector);
+
+    [StructLayout(LayoutKind.Explicit, Size = 8)]
+    public struct NamePlateColors {
+        [FieldOffset(0), CExportIgnore] public ulong Data;
+        [FieldOffset(0)] public ByteColor BackdropColor;
+        [FieldOffset(4)] public ByteColor FillColor;
```
wolfcomp:
```suggestion
        [FieldOffset(0)] public ByteColor EdgeColor;
        [FieldOffset(4)] public ByteColor Color;
```

To match the ATK system fully

Haselnussbomber: It's more so it matches the ConfigOptions as seen in the comment on `GetNamePlateColorType()` above, for example NamePlate**Edge**Party and NamePlate**Color**Party. 🙂
But Edge is probably just short for EdgeColor, like how the macro code is called.

Exter-N: If you prefer that, there you go, but these are used to populate fields in `Hud2NumberArray`. If their naming differs, I'm adding `<seealso />`s in regard to that.

PR #1564, opened 2025-08-24
Add a struct for CharacterBase.TempSlotData

Exter-N:


wolfcomp:
would be nice to have some idea what the different areas of the struct is even if its temporary

PR #1566, opened 2025-08-25
Fully cleanup enums before recreating

caitlyn-gg:
This change is primarily targeting IDA 9.x+ where duplicate enum member names can result in lingering members during cleanup, but I've also included an implementation for 7.x and 8.x for functional parity.

I've also reverted the back-to-front experiment as it did not improve performance on 9.x (WTF did you do, Hex-Rays?)

### DIFF
--- a/ida/ida_wrapper.py
+++ b/ida/ida_wrapper.py
@@ -599,8 +599,8 @@ class IdaInterface(BaseIdaInterface):
                 while mask != idaapi.DEFMASK:
                     masks.append(mask)
                     mask = idc.get_next_bmask(eid, mask)
-            members = []
 
+            members = []
             for mask in masks:
                 mem_val = idc.get_first_enum_member(eid, mask)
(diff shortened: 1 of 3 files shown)


ida/ida_wrapper.py
```diff
@@ -589,6 +586,39 @@ def remove_enum_member(self, eid: int, value: str, name: str):
...
+            for cid in members:
+                ok = ida_enum.del_enum_member(
+                    eid,
+                    ida_enum.get_enum_member_value(cid),
+                    ida_enum.get_enum_member_serial(cid),
+                    ida_enum.get_enum_member_bmask(cid)
+                )
+                
+                if not ok:
+                    raise RuntimeError("Failed to delete enum member {} in enum {}"%(cid, eid))
```
wolfcomp: Don't raise errors on IDA 7/8 in the wrapper

PR #1584, opened 2025-08-27
Add ShouldLimitFPS to UIModule

AtmoOmen:


AtmoOmen:
> This function already exists as vfunc in [UIModuleInterface](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/UI/UIModuleInterface.cs#L177).

Thank you Hasel

PR #1608, opened 2025-09-25
Additions

MidoriKami:
I'm 95% sure the return of LoadTexture is AtkTextureResource*, IDA shows a weird type for the return, but the actual object returned is AtkTextureResource*.

UnloadTexture returns 0xFFFFFFFF if given a null texture, but returns 0 in all other cases.
Not sure if long is the best return type for this, though it's the most accurate.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkInputManager.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkInputManager.cs
@@ -13,6 +13,8 @@ public unsafe partial struct AtkInputManager {
 
     [FieldOffset(0x80), FixedSizeArray] internal FixedSizeArray256<FocusEntry> _focusList;
 
+    [FieldOffset(0x1880)] public AtkCollisionNode* FocusedCollisionNode;
+
     [MemberFunction("E8 ?? ?? ?? ?? 33 C0 48 89 86 ?? ?? ?? ?? 88 86 ?? ?? ?? ?? 38 86")]
     public partial void HandleInput(AtkUnitManager* unitManager, AtkCollisionManager* collisionManager);
 
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkTextureResourceManager.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkTextureResourceManager.cs
@@ -1,3 +1,4 @@
+using FFXIVClientStructs.FFXIV.Client.System.Resource.Handle;
 using FFXIVClientStructs.FFXIV.Component.Exd;
 
 namespace FFXIVClientStructs.FFXIV.Component.GUI;
@@ -13,6 +14,14 @@ public unsafe partial struct AtkTextureResourceManager {
     // Icon ranges are inclusive: id ∈ [a,b]
     [FieldOffset(0x30), FixedSizeArray] internal FixedSizeArray4<StdPair<uint, uint>> _localizedIconRange;
     [FieldOffset(0x50)] public uint LocalizedIconRangeCount;
+
+    /// <summary>
+    /// Unloads the texture if the reference count is 1, otherwise decrements the textures reference count.
+    /// </summary>
+    /// <param name="textureHandle">Pointer to a texture resource</param>
+    /// <returns>-1 if the textureHandle is null, 0 otherwise</returns>
+    [MemberFunction("E8 ?? ?? ?? ?? 32 C0 45 84 FF")]
+    public partial long UnloadTexture(TextureResourceHandle* textureHandle);
 }
 
 public enum IconSubFolder {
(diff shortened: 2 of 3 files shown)


FFXIVClientStructs/FFXIV/Component/GUI/AtkTextureResourceManager.cs
```diff
@@ -13,6 +14,20 @@ public unsafe partial struct AtkTextureResourceManager {
...
     [FieldOffset(0x30), FixedSizeArray] internal FixedSizeArray4<StdPair<uint, uint>> _localizedIconRange;
     [FieldOffset(0x50)] public uint LocalizedIconRangeCount;
+
+    /// <summary>
+    /// Unloads the texture if the reference count is 1, otherwise decrements the textures reference count.
+    /// </summary>
+    /// <param name="textureHandle">Pointer to a texture resource</param>
+    /// <returns>-1 if the textureHandle is null, 0 otherwise</returns>
+    [MemberFunction("E8 ?? ?? ?? ?? 32 C0 45 84 FF")]
+    public partial long UnloadTexture(TextureResourceHandle* textureHandle);
```
Haselnussbomber: Looks like it's an int. Shows `0xFFFFFFFF` (equals -1) and writes to `eax`. 
👉 https://cs.brown.edu/courses/cs033/docs/guides/x64_cheatsheet.pdf
```suggestion
    public partial int UnloadTexture(TextureResourceHandle* textureHandle);
```

FFXIVClientStructs/FFXIV/Component/GUI/AtkTextureResourceManager.cs
```diff
@@ -13,6 +14,20 @@ public unsafe partial struct AtkTextureResourceManager {
...
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 8B F0 48 85 C0 74 35 48 8B 48 08")]
+    public partial AtkTextureResource* LoadIconTexture(int iconId, IconSubFolder folder);
+    
+    [MemberFunction("E8 ?? ?? ?? ?? 4C 8B E8 48 85 C0 74 52")]
+    public partial AtkTextureResource* LoadTexture(CStringPointer path, IconSubFolder folder);
```
Haselnussbomber: - Missing the attribute to generate string overloads.
- Second argument is saved in `Version`, which is the texture scale. If not 1 (means it's 2, because scale 0 doesn't exist), it inserts `_hr1` to the path.

```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? 4C 8B E8 48 85 C0 74 52"), GenerateStringOverloads]
    public partial AtkTextureResource* LoadTexture(CStringPointer path, int textureScale);
```

Haselnussbomber (changes requested):

FFXIVClientStructs/FFXIV/Component/GUI/AtkTextureResourceManager.cs
MidoriKami: Ahh yeah forgot the attribute, thanks for the param correction.

FFXIVClientStructs/FFXIV/Component/GUI/AtkTextureResourceManager.cs
MidoriKami: Oh whoops, not sure why I thought that 0xFFFFFFFF was bigger than 32bits. Thanks!

FFXIVClientStructs/FFXIV/Component/GUI/AtkInputManager.cs
```diff
@@ -13,6 +13,8 @@ public unsafe partial struct AtkInputManager {
 
     [FieldOffset(0x80), FixedSizeArray] internal FixedSizeArray256<FocusEntry> _focusList;
 
+    [FieldOffset(0x1880)] public AtkCollisionNode* FocusedCollisionNode;
```
Haselnussbomber: Doesn't look like a CollisionNode. Says AtkTextNode. 🤔
When I open windows it's a CollisionNode. Should probably be a AtkResNode*.

Image

wolfcomp: 100% is `AtkResNode*` as `Component::GUI::AtkInputManager_HandleFocus` checks if `a3[1]` is the same value to return early and `a3[1]` is set in `Component::GUI::AtkInputManager_SetFocus` as `AtkResNode*` from `a2`

MidoriKami: Ah alright, I was testing with KTK stuff, and everything that was responding to events was a Collision node. I didn't test with other node types. Thanks!

PR #1612, opened 2025-09-29
Add BitArray support to FixedSizeArray

Haselnussbomber:
We recently added the [`Span.CheckBitInSpan`](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/Interop/SpanExtensions.cs#L35-L64) extension function, but when I worked on it, I found [`System.Collections.BitArray`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/BitArray.cs) and wanted to make a generator for the bit arrays in UIState and PlayerState.

This is an addition to the FixedSizeArray generator.
When given `isBitArray: true` and a `bitCount`, it generates a public property `{FieldName}BitArray` that returns a `new BitArray(ptr, bitCount)`, which is similar to `Span<T>(ptr, length)`, that has a simple indexer to get or set the state of a bit in the array.

The analyzer checks if the bit count matches the size (`size == (bitCount + 7) / 8`), that the generic type is `byte`, and that `isString` is `false`.

Example:
```cs
[GenerateInterop]
[StructLayout(LayoutKind.Explicit, Size = 0x1A230)]
public unsafe partial struct UIState {
    [FieldOffset(0x19F1E), FixedSizeArray(isBitArray: true, bitCount: 1380)] internal FixedSizeArray173<byte> _seenCutscenes;
}
```
will generate
```cs
[global::System.Runtime.InteropServices.StructLayoutAttribute(global::System.Runtime.InteropServices.LayoutKind.Sequential, Pack = 1)]
[global::System.Runtime.CompilerServices.InlineArrayAttribute(173)]
internal struct FixedSizeArray173<T> where T : unmanaged
{
    private T _element0;
}

unsafe partial struct UIState
{
    /// <inheritdoc cref="_seenCutscenes" />
    [global::System.Diagnostics.CodeAnalysis.UnscopedRefAttribute] public Span<byte> SeenCutscenes => _seenCutscenes;
    /// <inheritdoc cref="_seenCutscenes" />
    public BitArray SeenCutscenesBitArray => new BitArray((byte*)global::System.Runtime.CompilerServices.Unsafe.AsPointer(ref _seenCutscenes[0]), 1380);
}
```
that can be used as
```cs
// ------
// properties for convenience
// ------

var pointer = uiState->SeenCutscenesBitArray.Pointer;
var bitCount = uiState->SeenCutscenesBitArray.BitCount;
var byteLength = uiState->SeenCutscenesBitArray.ByteLength;

// ------
// getter
// ------

var hasSeenCutscene1337 = uiState->SeenCutscenesBitArray[1337];

// or

uiState->SeenCutscenesBitArray.Get(1337);

// ------
// setter
// ------

// uiState->SeenCutscenesBitArray[1337] = true;
// will result in "Cannot modify the return value of 'UIState.SeenCutscenesBitArray' because it is not a variable",
// so we have to do:

var bitArray = uiState->SeenCutscenesBitArray;
bitArray[1337] = false;

// or

uiState->SeenCutscenesBitArray.Set(1337, false);

// ------
// iterator
// ------

foreach ((var index, var value) in uiState->SeenCutscenesBitArray) {
    ImGui.Text($"[{index}] = {value}");
}
```

(Note: I have not added any BitArrays to UIState yet.)

Added new Tests, which are passing:

Tests


Haselnussbomber:
Decided to rework this PR. It now reuses and extends FixedSizeArray(Attribute) instead.

Tests

PR #1628, opened 2025-10-19
Additions

MidoriKami:
Really small stuff, just some observations

The collision bit note is a little wordy, improved wording is welcomed if you have any suggestions

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkResNode.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkResNode.cs
@@ -74,7 +74,7 @@ public unsafe partial struct AtkResNode : ICreatable {
     /// <term>Bit 21 [0x100000]</term> Change CursorType to Clickable on hover<br/>
     /// <term>Bit 23 [0x400000]</term> Change CursorType to TextInput on hover<br/>
     /// <term>Bit 24 [0x800000]</term> Use elliptical collision instead of rectangular<br/>
-    /// <term>Bit 25 [0x1000000]</term> Use transformed collision instead of rectangular, if the object is rotated represents a strict collision check, instead of just bounding box
+    /// <term>Bit 25 [0x1000000]</term> Perform a stricter collision check. Instead of using the axis-aligned bounding box, the node's rotation and scale is taken into account.
     /// </summary>
     [FieldOffset(0xA0)] public uint DrawFlags;
 


FFXIVClientStructs/FFXIV/Component/GUI/AtkResNode.cs
```diff
@@ -72,7 +73,8 @@ public unsafe partial struct AtkResNode : ICreatable {
     /// <term>Bits 10-17</term> ClipCount<br/>
     /// <term>Bit 21 [0x100000]</term> Change CursorType to Clickable on hover<br/>
     /// <term>Bit 23 [0x400000]</term> Change CursorType to TextInput on hover<br/>
-    /// <term>Bit 24 [0x800000]</term> Use elliptical collision instead of rectangular
+    /// <term>Bit 24 [0x800000]</term> Use elliptical collision instead of rectangular<br/>
+    /// <term>Bit 25 [0x1000000]</term> Use transformed collision instead of rectangular, if the object is rotated represents a strict collision check, instead of just bounding box
```
redstrate:
```suggestion
    /// <term>Bit 25 [0x1000000]</term> Perform a stricter collision check. Instead of using the axis-aligned bounding box, the node's rotation and scale is taken into account.
```

I don't know if scale is actually taken into account, but I think it balances out the sentence a bit if you want to include it!

MidoriKami: I can confirm that it does account for scale.

FFXIVClientStructs/FFXIV/Component/GUI/AtkImageNode.cs
```diff
@@ -36,6 +36,7 @@ public void Ctor() {
 public enum ImageNodeFlags {
     FlipH = 0x01,
     FlipV = 0x02,
+    RenderMiniMap = 0x04, // Renders minimap using the set textures as a mask
```
wolfcomp, line 39: This flag has nothing to do with the minimap and is a different type of flag that just happens to be set on the minimap.

MidoriKami: Do you have any examples, or information to backup that claim?

MidoriKami: And failing that, what exactly are you requesting it be changed to?

MidoriKami:
If anyone wants to attempt to more properly define the image flags you can see this discussion in the discord: https://discord.com/channels/581875019861328007/1246473943548825730/1430687989222211735

And this data dump for various flag combinations: https://discord.com/channels/581875019861328007/1246473943548825730/1430764101939888190

Contrary to wolfcomp's assertion, I have seen no information to indicate that these flags are not hardcoded to the minimap.

PR #1629, opened 2025-10-26
Add FishingEventHandler

Knightmore:


FFXIVClientStructs/FFXIV/Client/Game/Event/FishingEventHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+using ValueType = FFXIVClientStructs.FFXIV.Component.GUI.ValueType;
+
+namespace FFXIVClientStructs.FFXIV.Client.Game.Event;
+
+[GenerateInterop]
+[Inherits<EventHandler>]
+[StructLayout(LayoutKind.Explicit, Size = 0x28C)]
+public unsafe partial struct FishingEventHandler {
+	[FieldOffset(0x230)] private int _canFish;	
+	public bool CanFish => _canFish != 0;
```
Haselnussbomber: Runtime marshalling is disabled as per [csproj](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIVClientStructs.csproj#L8). You can just write this here:
```suggestion
	[FieldOffset(0x230)] public bool CanFish;
```

Haselnussbomber (changes requested):

FFXIVClientStructs/FFXIV/Client/Game/Event/FishingEventHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+	
+    [MemberFunction("48 89 5C 24 ?? 48 89 74 24 ?? 57 48 83 EC 30 49 8B F8 48 8B DA 48 8B F1 41 83 F9 02")]
+    public partial AtkValue* ChangeBait(AtkValue* returnValue, AtkValue* value, int arg0);
```
Haselnussbomber: Also, add `Inherits<AtkModuleInterface.AtkEventInterface>` after `Inherits<EventHandler>`, remove this function and use the inherited ReceiveEvent function instead.

PR #1635, opened 2025-11-05
Added Request Refresh on Operations Guide

Glyceri:
Unk19 is a boolean that requests a refresh of the Operations Guide.


Glyceri:
And the AttachedToAddon(2) is the addon the OperationsGuide is currently attached to.

PR #1640, opened 2025-11-09
Update AtkComponentTextInput stuff

Haselnussbomber:


FFXIVClientStructs/FFXIV/Component/GUI/AtkComponentInputBase.cs
```diff
@@ -10,15 +10,51 @@ namespace FFXIVClientStructs.FFXIV.Component.GUI;
...
+/// </summary>
+public enum InputCallbackType {
+    /// <summary> The user pressed the <c>Enter</c> key. </summary>
+    Enter = 0,
+    /// <summary> The user pressed the <c>Escape</c> key. </summary>
+    Escape = 1,
+    /// <summary> The user clicked outside the input area, causing it to lose focus. </summary>
+    FocusLost = 2,
+    /// <summary> The user modified the input text. </summary>
+    TextChanged = 3,
```
Haselnussbomber, line 45: I need to check this again later. This is InputBase, so it would also apply to NumberInputs, thus the name needs adjustment.

Haselnussbomber: I cannot find any reference to the Callback in AtkComponentNumericInput.ReceiveEvent, so I guess we can leave it like that.

PR #1647, opened 2025-11-25
Misc findings

Jaksuhn:


FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
```diff
@@ -139,6 +146,10 @@ public unsafe partial struct InventoryManager {
     /// <summary> Gets the number of (limited) tomestones the user has acquired during the current reset cycle. </summary>
     public int GetWeeklyAcquiredTomestoneCount() => GetLimitedTomestoneCount(GetSpecialItemId(9));
 
+    /// <remarks>Return value is a LogMessage Id. 0 if you can equip. Row pointer is optional</remarks>
+    [MemberFunction("E8 ?? ?? ?? ?? 85 C0 75 ?? 80 7E")]
+    public static partial int CanEquip(uint itemId, byte race, byte sex, ushort level, byte classJobId, byte grandCompany, byte pvpRank, ExcelRow* item);
```
Haselnussbomber: The item row pointer is incorrect. It points to the Data inside ExcelRow.
```suggestion
    public static partial int CanEquip(uint itemId, byte race, byte sex, ushort level, byte classJobId, byte grandCompany, byte pvpRank, [CExporterExcel("Item")] void* itemRow);
```

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentHUD.cs
```diff
@@ -75,6 +75,9 @@ public unsafe partial struct AgentHUD {
     [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 48 8B CF 4C 89 B4 24 B8 08 00 00")]
     public partial void OpenSystemMenu(AtkValue* atkValueArgs, uint menuSize);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 40 B5 ?? 41 B9")]
+    public partial bool ExecuteSystemMenu(int menuType, int commandId, int a4 = -1, byte* a5 = null);
```
Haselnussbomber: Hm, I don't think this is correct.
From what I can see this function refreshes the MainCross or MainCommand addons. Would have to look into those addons to figure out what for.
In any case, it doesn't seem to have anything to do with executing SystemMenus (that's the Esc menu btw).
Were you looking for `UIModule.ExecuteMainCommand`?

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentReturn.cs
```diff
@@ -0,0 +1,19 @@
+namespace FFXIVClientStructs.FFXIV.Client.UI.Agent;
+
+// Client::UI::Agent::AgentReturn
+//   Client::UI::Agent::AgentInterface
+//     Component::GUI::AtkModuleInterface::AtkEventInterface
+[Agent(AgentId.Return)]
+[GenerateInterop]
+[Inherits<AgentInterface>]
+[StructLayout(LayoutKind.Explicit, Size = 0x60)]
+public unsafe partial struct AgentReturn {
```
Haselnussbomber, line 10: Isn't this agent unnecessary? You can just use ActionManager directly instead. Unless you want to hook it, I guess.

To check if you can cast Return:
`ActionManager.Instance()->GetActionStatus(ActionType.Action, 6) == 0`

To cast Return:
`ActionManager.Instance()->UseAction(ActionType.Action, 6)`

FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
```diff
@@ -137,6 +137,9 @@ public unsafe partial struct RaptureAtkModule {
     [VirtualFunction(63), GenerateStringOverloads]
     public partial bool OpenMapWithMapLink(CStringPointer mapLink);
 
+    [MemberFunction("48 89 5C 24 ?? 48 89 74 24 ?? 48 89 7C 24 ?? 55 41 56 41 57 48 8B EC 48 83 EC 40 4C 8B F1")]
+    public partial void HandleItemMove(void* outValue, AtkValue* atkValues);
```
Haselnussbomber: Please use the arguments and return type of [`AtkModule.CallbackHandlerDelegate`](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Component/GUI/AtkModule.cs#L18):

```suggestion
    public partial AtkValue* HandleItemMove(AtkValue* returnValue, AtkValue* values, uint valueCount);
```

FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
```diff
@@ -63,6 +64,12 @@ public unsafe partial struct GameMain {
     [MemberFunction("E8 ?? ?? ?? ?? 80 63 50 FE")]
     public partial void SetActiveFestivals(uint festival1, uint festival2, uint festival3, uint festival4); // Applies immediately regardless of client state
 
+    [MemberFunction("E8 ?? ?? ?? ?? 8D 46 0A")]
+    public static partial nint ExecuteCommand(int command, int param1 = 0, int param2 = 0, int param3 = 0, int param4 = 0);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 8B 7B 08 45 33 C0")]
+    public static partial nint ExecuteLocationCommand(int command, Vector3 position, int param1 = 0, int param2 = 0, int param3 = 0, int param4 = 0);
```
Haselnussbomber: - Both of these return bool.
- Vector3 on ExecuteLocationCommand needs to be passed as pointer.
- Incorrect signature for ExecuteLocationCommand.

```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? 8D 46 0A")]
    public static partial bool ExecuteCommand(int command, int param1 = 0, int param2 = 0, int param3 = 0, int param4 = 0);

    [MemberFunction("E8 ?? ?? ?? ?? 49 8D 54 24 ?? B9")]
    public static partial bool ExecuteLocationCommand(int command, Vector3* position, int param1 = 0, int param2 = 0, int param3 = 0, int param4 = 0);
```

Haselnussbomber (changes requested):

FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
Jaksuhn: Ah, my bad. Overlooked it after I saw that it doesn't even seem to be used (well, rather, there's a check in the beginning of the function for it but I never got it to trigger in testing, even just passing null. I didn't really understand it)

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentHUD.cs
Jaksuhn: From my understanding it seems to be a combination based on `menuType`. That being said yes despite that I did just call it ExecuteSystemMenu even though it's probably multipurpose (I didn't look much further, I only cared for system menu). 

It returns the MainCommandId of the system menu item selected, at least for `menuType == 1`. CommandId is a bitmask? in other cases.

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentReturn.cs
Jaksuhn: Yeah I was hooking it to track returns. Not the most useful agent, I agree. It's at least got a whole 1 field! (there's at least a couple more: some container? at `0x30`, and a static int initialised to 1 at `0x58`)
Is the redundancy a particular problem? I'm not sure if there's really rules/guidelines for what's out of scope or not

FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs
Jaksuhn: Thank you! Was unaware of this

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentHUD.cs
Jaksuhn: Maybe something like this would be better? `commandId` might not be a perfect name but overall a bit more accurate.

```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? 40 B5 ?? 41 B9")]
    public partial bool RefreshMainCommandAddon(RefreshMainCommandAddonOperation operation, int commandId, int a4 = -1, byte* a5 = null);
    
    enum RefreshMainCommandAddonOperation {
          InitializeMainCommand = 0, // sets up categories for _MainCommand
          NotifyCommandSelected = 1, // Notifies _MainCommand of selected command (e.g. from system menu)
          UpdateCommandState = 2, // Update _MainCommand/_MainCross
          Unk4 = 4,
          Unk5 = 5,
          Unk6 = 6,
          Unk7 = 7, // Sends enabled commands to _MainCommand/_MainCross?
};
```

Jaksuhn:
(I was not sure if you get notifications for review replies or only when the review is re-requested)

FFXIVClientStructs/FFXIV/Client/Game/Character/Character.cs
```diff
@@ -11,6 +11,7 @@ namespace FFXIVClientStructs.FFXIV.Client.Game.Character;
 [StructLayout(LayoutKind.Explicit, Size = 0x2360)]
 [VirtualTable("48 8D 05 ?? ?? ?? ?? 48 89 07 48 8D 8F ?? ?? ?? ?? 48 8D 05 ?? ?? ?? ?? 48 89 87 ?? ?? ?? ?? E8 ?? ?? ?? ?? 48 8D 8F ?? ?? ?? ?? 33 ED 48 8D 05 ?? ?? ?? ??", 3)]
 public unsafe partial struct Character {
+    [FieldOffset(0x05FC)] public bool IsFlying;
```
Haselnussbomber: This does not seem to be a bool:
`(unsigned __int8)(g_Client::Game::Control::Control_Instance.LocalPlayer->field_5FC - 1) > 2u`
Seen at 0x140822530 and other addresses.

FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
```diff
@@ -87,6 +88,12 @@ public unsafe partial struct InventoryManager {
     [MemberFunction("E8 ?? ?? ?? ?? 8B F8 39 43 78")]
     public partial uint GetRetainerGil();
 
+    [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? BE ?? ?? ?? ?? 48 8D 45 90 8B CE 44 89 30")]
+    public partial int MoveFromRetainerMarketToPlayerInventory(InventoryType srcInv, ushort srcSlot, uint quantity);
```
Haselnussbomber, line 92: Wrong signature?

0x140841150 moves the item from RetainerMarket to the retainers inventory.
0x140841310 moves the item from RetainerMarket to the players inventory (sig is not resolving to this).

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentHUD.cs
Haselnussbomber: Alright, I looked into this.

- `param1` is uint.
- No call passes anything to `param2` or `param3` (string pointers) from what I could see, so we leave them as is.
- MainCommandOperation 3 is completely unused (no call with it, no case in the MainCommand addon), so I left it out.
- The only thing having to do with SystemMenu is operation 4, which closes a MainCommand category menu and opens the SystemMenu.
- Couldn't figure out what 6 does. Something with VisibilityFlags for the menu? Anyway, can stay unk.

If you use this suggestion, please also update the name in the data.yml and move the enum beneath the other functions. 🙂

```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? 40 B5 ?? 41 B9")]
    public partial bool HandleMainCommandOperation(MainCommandOperation operation, uint param1, int param2 = -1, byte* param3 = null);

    public enum MainCommandOperation {
        SetCategoryNames = 0,
        /// <remarks> param1: MainCommandId </remarks>
        ExecuteMainCommand = 1,
        /// <remarks> param1: MainCommandId | 0x10000 for enabled, otherwise just MainCommandId. </remarks>
        SetMainCommandEnabledState = 2,

        /// <remarks> Fired when <see cref="InputId.MENU_SYSTEM"/> was pressed. </remarks>
        OpenSystemMenu = 4,
        /// <remarks> param1: 0 = Hide, 1 = Show </remarks>
        SetVisibility = 5,
        // Unk6 = 6,
        SetCategoryPatchMarks = 7,
    }
```

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentReturn.cs
Haselnussbomber: > Is the redundancy a particular problem?

Not to my knowledge.

FFXIVClientStructs/FFXIV/Client/Game/InventoryManager.cs
Jaksuhn: God did I just misclick? That one was from updating idbs. Sorry for the blatantly messed up problems like this
(also thanks for 0x140841150, that's cool)

FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentHUD.cs
Jaksuhn: Went ahead and used this since I'm terrible with names and these sound about right from what I saw. Very much appreciate the extra RE effort cause some of those flags I did _not_ get at all, like 6/7.

(also fixed the enum position and data.yml)

FFXIVClientStructs/FFXIV/Client/Game/Character/Character.cs
Jaksuhn:
```suggestion
    [FieldOffset(0x5FC)] public MovementStateOptions MovementState;
    
    public enum MovementStateOptions {
        /// <remarks>Includes swimming above water and jumping/falling</remarks>
        Grounded = 0,
        Flying = 1,
        /// <remarks>Being mounted doesn't matter</remarks>
        Underwater = 2,
   }
```

Not really sure about the `>2` values. Seems to check those for camera related reasons, but can't think of any state that actually puts you in those.

(If this is fine I'll make a separate commit with the enum in the proper place. Shame you can't do that in  suggestions)

Haselnussbomber: Yeah, I don't know either. I have a feeling value 3 is PvP Spectator mode related, since I see a check for ParticipatingInCustomMatch, it has to do with targeting and Camera3, but I don't know how I could test that and I also do not care about that. 😅

I think we can add it like this, unless someone has something against it.
This is technically in the MovementContainer nobody dares to add for obvious reasons.

Jaksuhn: Yeah I saw the custom match bit too, but I'm only free trial at the moment so no way of going in to test.
Added it as is with the enum moved. Hopefully it being movement related isn't too big of a deal. If it is I can remove it ofc.

Haselnussbomber: > Includes swimming above water

What does that mean? Just swimming? Maybe the value 2 should be named Diving, like the condition?

Jaksuhn: Just when you're floating in the water, but haven't dove underwater. You're quite literally not "grounded" when in water but I don't really have a better term. "Not mounted" wouldn't work either since being mounted is also `0`. Also chocobo porters, jump pads and waterfalls (everything I could think of with forced movement) is all `0`.

Maybe "floating in water" is more clear

Haselnussbomber: I feel this should be more like this
```cs
public enum MovementStateOptions {
    Normal = 0,
    Flying = 1,
    Diving = 2,
    // Spectating = 3,
}
```

Jaksuhn: Mmm, yeah that's a lot better. I'll swap it to that

PR #1648, opened 2025-11-26
Stuff related to level object animations

xanunderscore:


xanunderscore:
Pohky led me down the path of the RTTI, so I'm gonna reopen this once I've fixed the class names

PR #1660, opened 2025-12-13
Apricot

chirpxiv:
apricot is a nice fruit

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Vfx/VfxResourceHandle.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Vfx/VfxResourceHandle.cs
@@ -2,8 +2,7 @@
 
 [StructLayout(LayoutKind.Explicit, Size = 8)]
 public struct VfxResourceHandle {
-    [FieldOffset(0)] public ulong Value;
-
-    [FieldOffset(0)] public uint Id;
-    [FieldOffset(4)] public uint Index;
+    [FieldOffset(0), CExporterUnion("Value")] public ulong Value;
+    [FieldOffset(0), CExporterUnion("Value")] public uint Id;
+    [FieldOffset(4), CExporterUnion("Value")] public uint Index;
 }


FFXIVClientStructs/Apricot/ApricotEngine.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    [MemberFunction("E8 ?? ?? ?? ?? 4B 8B 0C 27")]
+    public static partial ApricotEngine* GetInstance();
+    
+    [GenerateInterop(isInherited: true)]
+	[StructLayout(LayoutKind.Explicit, Size = 0x83030)]
+	public partial struct DataContainer {
+        [FieldOffset(0x2030), FixedSizeArray] internal FixedSizeArray2048<InstanceContainer> _instances;
+
+        public DocumentInstance* GetDocument(int index) => this._instances[index].Instance;
```
Haselnussbomber, line 19: ~~This isn’t needed. We have a GetPointer extension.~~
Nevermind, misread, sorry.

FFXIVClientStructs/Apricot/Data/RgbFunctionCurve.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Numerics;
+
+namespace FFXIVClientStructs.Apricot.Data;
+
+[StructLayout(LayoutKind.Explicit)]
+public struct RgbFunctionCurve {
+	[FieldOffset(0x0)] public uint Data;
+    // not a fixed size array, not sure how else to handle this?
+    [FieldOffset(0x4), FixedSizeArray] internal unsafe fixed byte _keys[0x10];
```
Haselnussbomber, line 9: I think what you were looking for was `FixedSizeArray2<Pointer<RgbKey>>`.
But in this case, with a dynamic sized array, it probably won’t help much.

wolfcomp: Could do
```cs
[FieldOffset(0x4)] public RgbKey* KeyPointer;

public Span<RgbKey> Keys => new Span(KeyPointer, KeyCount);
```
with looking at how the `GetKey` function works

FFXIVClientStructs/FFXIV/Client/Graphics/Vfx/VfxResourceHandle.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace FFXIVClientStructs.FFXIV.Client.Graphics.Vfx;
+
+[StructLayout(LayoutKind.Explicit, Size = 8)]
+public struct VfxResourceHandle {
+    [FieldOffset(0)] public ulong Value;
```
wolfcomp: Just see where there are union fields already used to find how to utilize it.

`InventoryItem` is a good example for this use case.

CExporter/Exporter.cs
Caraxi, line 118: Is this `Debugger.Break()` intentionally left here for some reason?

chirpxiv: this was from rebasing incorrectly, resolved now

FFXIVClientStructs/Apricot/Data/RgbFunctionCurve.cs
chirpxiv: > I think what you were looking for was `FixedSizeArray2<Pointer<RgbKey>>`. But in this case, with a dynamic sized array, it probably won’t help much.

Originally tried this with a `FixedSizeArray1`, but it throws a `IndexOutOfRangeException` beyond the first element.



> Could do
> 
> ```cs
> [FieldOffset(0x4)] public RgbKey* KeyPointer;
> 
> public Span<RgbKey> Keys => new Span(KeyPointer, KeyCount);
> ```
> 
> with looking at how the `GetKey` function works

There isn't a pointer here, it's an inline array with a dynamic size.

FFXIVClientStructs/FFXIV/Client/Graphics/Vfx/VfxResourceHandle.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace FFXIVClientStructs.FFXIV.Client.Graphics.Vfx;
+
+[StructLayout(LayoutKind.Explicit, Size = 8)]
+public struct VfxResourceHandle {
+    [FieldOffset(0), CExporterUnion("Value")] public ulong Value;
+    [FieldOffset(0), CExporterUnion("Value")] public uint Id;
+    [FieldOffset(4), CExporterUnion("Value")] public uint Index;
```
wolfcomp, line 7: this will export the unions as
```cpp
union Value {
  uint64_t Value;
  uint32_t Id;
  uint32_t Index;
}
``` 
all which are at field offset 0

PR #1678, opened 2025-12-20
Map player attribute setters used during packet processing

redstrate:
These are the two functions used in HandlePlayerStatsPacket, and each one takes a different kind of index.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -2710,7 +2710,9 @@ classes:
       0x140B00170: GetCurrentClassJobExp
       0x140B00220: GetClassJobNeededExp
       0x140B00290: GetCurrentClassJobNeededExp
+      0x140B00360: SetMainAttributeByIndex # (this, index, value)
       0x140B00380: GetAttributeByIndex # (this, index)
+      0x140B003A0: SetAttributeByIndex # (this, index, value)
       0x140AFD850: SetEurekaRank
       0x140AFD950: SetBozjaRank
       0x140AFDE00: TrackTraitUnlock


ida/data.yml
```diff
@@ -2697,7 +2697,9 @@ classes:
       0x140B00170: GetCurrentClassJobExp
       0x140B00220: GetClassJobNeededExp
       0x140B00290: GetCurrentClassJobNeededExp
+      0x140B00360: SetNonMainAttributeByIndex # (this, index, value)
```
redstrate: Meh not in love with this name but didn't know what to call it:

```c
void Client::Game::UI::PlayerState.SetNonMainAttributeByIndex(PlayerState *param_1,int param_2,int param_3)
{
  if (param_2 - 1U < 6) {
    param_1->Attributes[(ulonglong)(param_2 - 1U) - 6] = param_3;
  }
  return;
}
```

It looks like it skips over the main attributes

Haselnussbomber: No, the opposite. It checks < 6 and sets the base values for the main attributes(?). The fields should be an array.

```c
void __fastcall sub_140B00360(Client::Game::UI::PlayerState *a1, int a2, int a3)
{
  __int64 v3; // rdx

  v3 = (unsigned int)(a2 - 1);
  if ( (unsigned int)v3 <= 5 )
    *(&a1->BaseStrength + v3) = a3;
}
```

In your version it access the array Attributes and subtracts 6 to get to the correct address.

redstrate: Yeah you're right, I dunno why the packet is order that way but it is!

PR #1683, opened 2025-12-22
LogMessage Queue

RedworkDE:
Everything that is useful to replace using regex for parsing the various system messages in the chat by simple id comparisons.

I am still working on a wrapper in dalamud to make this actually easy to use.

The last commit with the `EntityRelationKind` enum is kinda eh because of the existing `GetLogMessageDetail` method which returns this info with different values, making the design somewhat awkward, but the current version should be fully compatible for existing users. Alternatively we could just drop that commit, ditch the enum for explaining the values and maybe add a comment like on the existing `GetLogMessageDetail` instead.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureLogModule.cs
+++ b/FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureLogModule.cs
@@ -208,6 +208,30 @@ public struct RaptureLogModuleTab {
     [FieldOffset(0x68)] public Utf8String VisibleLogLines;
 }
 
+[GenerateInterop]
+[StructLayout(LayoutKind.Explicit, Size = 0x248)]
+public partial struct LogMessageQueueItem {
+    [FieldOffset(0)] public StdDeque<TextParameter> Parameters;
+    [FieldOffset(0x28), FixedSizeArray(isString: true)] internal FixedSizeArray256<byte> _sourceName;
+    [FieldOffset(0x128), FixedSizeArray(isString: true)] internal FixedSizeArray256<byte> _targetName;
+    [FieldOffset(0x228)] public uint LogMessageId;
+    [FieldOffset(0x22C)] public EntityRelationKind SourceKind;
+    [FieldOffset(0x22D)] public EntityRelationKind TargetKind;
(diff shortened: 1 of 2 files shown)


FFXIVClientStructs/FFXIV/Client/UI/Misc/RaptureLogModule.cs
```diff
@@ -69,6 +84,24 @@ public unsafe partial struct RaptureLogModule {
...
+    public partial void ShowLogMessageSourceObjectTargetObject(uint logMessageId, GameObject* source, GameObject* target);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 44 8B 45 DF")]
+    public partial void ShowLogMessageSourceObjectTargetObjectTextParameters(uint logMessageId, GameObject* source, GameObject* target, StdDeque<TextParameter>* value);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 44 8B 4D 18"), GenerateStringOverloads]
+    public partial void ShowLogMessageSourceTarget(uint logMessageId, CStringPointer sourceName, byte sourceGender, CStringPointer targetName, byte targetSex, ushort sourceHomeWorld, ushort targetHomeWorld);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 4D 85 FF 74 3F"), GenerateStringOverloads]
+    public partial void ShowLogMessageSourceTargetTextParameters(uint logMessageId, CStringPointer sourceName, byte sourceSex, CStringPointer targetName, byte targetSex, StdDeque<TextParameter>* value, ushort sourceHomeWorld, ushort targetHomeWorld);
```
Haselnussbomber: Do we actually need those? Our StdDeque implementation is readonly, so you cannot create your own parameters. Not sure if those are useful with this limitation.

Haselnussbomber, line 100: Generally fine to add, but note that our StdDeque implementation is readonly, so you cannot create your own parameters. It's generally advised to use `ISeStringEvaluator.EvaluateFromLogMessage` in Dalamud and printing the result to Chat or whatever, if needed.

PR #1689, opened 2025-12-30
Update AgentRepair.cs

Knightmore:


Knightmore:
I don't think that we can get around the last breaking change, as I added an enum so users know right away what the possible values are and don't set any bogus value.

MidoriKami:
You can by doing the following:
```cs
   [FieldOffset(0x5C), Obsolete("Use InventoryContainer field")] public int InventoryContainerIndex; // Used to lookup static array. Mapped index ids of repair gear dropdown. (7 = Equipped, 0 = Main/Off Hand, 1 = Head/Body/Hands, ...)
   [FieldOffset(0x5C)] public InventoryContainer InventoryContainer;
```

Haselnussbomber:
Worked it out a bit more and made it a bit cleaner. :)

PR #1705, opened 2026-01-18
Add more bitarrays for special and misc content

redstrate:
When I originally added the content unlock & completion bitmasks, there were a few omissions like Rival Wings and V&C Dungeons. For some reason, these are stored a bit differently.

Some content can override which array is used, either in the InstanceContentType or InstanceContent sheet. Older content like Rival Wings uses the former, and newer content like V&C Dungeons prefers the latter. That's the meaning behind the "specialContent" and "miscContent" arrays. I added some remarks detailing which content is stored in which, since the names are very broad.

I also added arrays for BeginnerTraining and MaskedCarnivale content.


redstrate:
Relevant EXDSchema change to help understand what the client is doing: https://github.com/xivdev/EXDSchema/pull/158

PR #1716, opened 2026-01-30
Correct call for Human.UpdateDrawData function with new struct created in stack of functions

wolfcomp:
Depends on: #1725

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -7280,7 +7280,7 @@ classes:
       0x1404320D0: SetupVisor
   Client::Graphics::Scene::Human::DrawData:
     funcs:
-      0x14042D980: ctor
+      0x14042C5C0: ctor
   Client::Graphics::Scene::Demihuman:
     vtbls:
       - ea: 0x142097610


FFXIVClientStructs/FFXIV/Client/Graphics/Scene/Human.cs
```diff
@@ -82,13 +84,29 @@ public ConstantBufferPointer<Vector4> DecalColorTypedCBuffer
...
+    public partial struct DrawData : ICreatable
+    {
+        [FieldOffset(0x00)] public CustomizeData CustomizeData;
+        [FieldOffset(0x1A)] public byte AnimationVariant;
+        [FieldOffset(0x1E)] private ushort Unk1E;
+        [FieldOffset(0x20), FixedSizeArray] internal FixedSizeArray10<EquipmentModelId> _equipments;
+        [FieldOffset(0x70), FixedSizeArray] internal FixedSizeArray2<EquipmentModelId> _glasses;
+
+        [MemberFunction("66 0F 6F 0D ?? ?? ?? ?? 0F 57 C0")]
+        public partial void Ctor();
```
Haselnussbomber: Does that not return `this`?
```cs
.text:000000014042DD0E 48 8B C1                                mov     rax, rcx
.text:000000014042DD11 C3                                      retn
```
Not that it's used anywhere, but still.

wolfcomp: Yes so is basically every other `Ctor` where `ICreatable` has also been used for the most part as with the MSVC compiler option all of them have the return of `T*`

PR #1727, opened 2026-02-18
Adds lots of stuff to Scene objects and adds Layout level staining support.

universalconquistador:
- Adds `Common::Math::SphereBounds`, `Common::Math::OrientedBounds`, and `Common::Math::AxisAlignedBounds` math types. I did not update the few places that use `Vector4` for sphere bounds instead of `SphereBounds`.

 - Adds vfuncs for Scene object updates, computing the 3 types of bounds, hit testing, working with attach bones, and dither transparency. Also added an enum for the Dtor mode so we can standardize that a bit. Tested the 3 bounds querying (some cases are wonky e.g. characters with oriented and aligned bounds, but I believe that's a sqenix problem), both hit test methods, and dither transparency (SE only implemented it for `BgObjects`). Did not test the attach bone stuff.

 - Adds the static Create helpers for `BgObject` and `VfxObject`, stain support for `BgObject`, and some VFX fields from OGT. Tested that I can create and destroy both, and set VFX color and `BgObject` stain.

 - Adds support for staining `SharedGroupLayoutInstance`s and in general via `ILayoutInstance.ApplyStain`. I have not tested these myself.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Render/Light.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Render/Light.cs
@@ -1,6 +1,3 @@
-using System;
-using System.Collections.Generic;
-using System.Text;
 using FFXIVClientStructs.FFXIV.Common.Math;
 
 namespace FFXIVClientStructs.FFXIV.Client.Graphics.Render;
@@ -10,7 +7,7 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Render;
 //     Client::Graphics::ReferencedClassBase
 [GenerateInterop]
 [Inherits<ReferencedClassBase>]
-[StructLayout(LayoutKind.Explicit, Size = 0x160)]
+[StructLayout(LayoutKind.Explicit, Size = 0x130)]
 public unsafe partial struct Light {
     /// <summary>
     /// A special <see cref="MaxRange"/> value that disables range checks.


FFXIVClientStructs/FFXIV/Client/Graphics/Scene/BgObject.cs
```diff
@@ -7,4 +8,45 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Scene;
 [StructLayout(LayoutKind.Explicit, Size = 0xE0)]
 public unsafe partial struct BgObject {
     [FieldOffset(0x90)] public ModelResourceHandle* ModelResourceHandle;
+
+    [FieldOffset(0xA0)] internal Matrix4x4* CachedTransformMatrices;
+    [FieldOffset(0xA8)] public BgObjectStainBuffer* StainBuffer;
+    [FieldOffset(0xB0)] internal Transform* CachedTransform;
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 89 43 30 48 8B D7")]
+    public static partial BgObject* Create(byte* modelGamePath, byte* pool, BgObject* existingAllocation);
```
Haselnussbomber: Using our generator will allow to pass strings directly to this method.
Since the last parameter can be optional, just default it to null here (nullptr).

```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? 48 89 43 30 48 8B D7"), GenerateStringOverloads]
    public static partial BgObject* Create(CStringPointer modelGamePath, CStringPointer pool, BgObject* existingAllocation = null);
```

FFXIVClientStructs/FFXIV/Client/Graphics/Scene/VfxObject.cs
```diff
@@ -13,4 +14,22 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Scene;
...
+    [FieldOffset(0x1B8)] public int StaticCaster;
+
+    [FieldOffset(0x1C0)] public int StaticTarget;
+
+    [FieldOffset(0x248)] internal byte SomeFlags;
+
+    [FieldOffset(0x260)] public Vector4 Color;
+
+    [MemberFunction("E8 ?? ?? ?? ?? F3 0F 10 35 ?? ?? ?? ?? 48 89 43 08")]
+    public static partial VfxObject* Create(byte* vfxGamePath, byte* pool);
```
Haselnussbomber:
```suggestion
    [MemberFunction("E8 ?? ?? ?? ?? F3 0F 10 35 ?? ?? ?? ?? 48 89 43 08"), GenerateStringOverloads]
    public static partial VfxObject* Create(CStringPointer vfxGamePath, CStringPointer pool);
```

FFXIVClientStructs/FFXIV/Client/LayoutEngine/Group/SharedGroupLayoutInstance.cs
```diff
@@ -36,4 +52,75 @@ public unsafe partial struct SharedGroupLayoutInstance {
...
+    /// <summary>
+    /// Looks up the sRGB color for the object stain choice at the given index.
+    /// </summary>
+    /// <remarks>
+    /// Index zero is a sentinel representing leaving the default stain, and will return empty black.
+    /// </remarks>
+    /// <param name="stainIndex">The object stain index to look up the color of.</param>
+    /// <returns>The sRGB color for the given object stain.</returns>
+    [MemberFunction("33 C0 81 F9 ?? ?? ?? ?? 0F 42 C1")]
+    public static partial ByteColor GetObjectStainColorByIndex(byte stainIndex);
```
Haselnussbomber: This returns a pointer: `&byte_1429A88BC`.

```suggestion
    public static partial ByteColor* GetObjectStainColorByIndex(byte stainIndex);
```

FFXIVClientStructs/FFXIV/Client/LayoutEngine/ILayoutInstance.cs
```diff
@@ -116,6 +117,9 @@ public unsafe partial struct Identifier {
     [VirtualFunction(55)]
     public partial bool WantToBeActive();
 
+    [VirtualFunction(58)]
+    public partial ByteColor ApplyStain(ByteColor* stainColorSrgb);
```
Haselnussbomber: Not sure that actually returns anything. Would have to look at the call sites, but it doesn't move anything into RAX register before returning.

Haselnussbomber (changes requested):
Just a quick glance at it.
It would also require you to add the functions and vfuncs to the data.yml once you're good with the names.

FFXIVClientStructs/FFXIV/Client/Graphics/Scene/VfxObject.cs
```diff
@@ -13,4 +14,22 @@ namespace FFXIVClientStructs.FFXIV.Client.Graphics.Scene;
...
+
+    [MemberFunction("E8 ?? ?? ?? ?? ?? ?? ?? 8B 4A ?? 85 C9")]
+    public partial uint Update(float deltaSeconds, uint flags = uint.MaxValue);
```
Haselnussbomber: I think "flags" is not really correct here. I don't know what it's used for, but it looks like it should be int (-1) as it checks `a3 >= 0` and `a2 <= 7` in the function it passes it to.

Also doesn't return anything.

FFXIVClientStructs/FFXIV/Client/LayoutEngine/ILayoutInstance.cs
pohky: it does not yeah (`1406dced2` and `1406e14da` if you want to check call sites)

ida/data.yml
```diff
@@ -7295,12 +7320,13 @@ classes:
         base: Client::Graphics::Scene::DrawObject
     funcs:
       0x140443640: ctor
-  Client::Graphics::Scene::BGObject:
+  Client::Graphics::Scene::BgObject:
```
wolfcomp:
```suggestion
  Client::Graphics::Scene::BGObject:
```

This is capitalized and the CS struct need to adjust to the naming it has in the RTTI info we have from some very ancient version of the game.

FFXIVClientStructs/FFXIV/Client/Graphics/Scene/Object.cs
```diff
@@ -12,6 +12,9 @@ public unsafe partial struct Object {
     [FieldOffset(0x20)] public Object* PreviousSiblingObject;
     [FieldOffset(0x28)] public Object* NextSiblingObject;
     [FieldOffset(0x30)] public Object* ChildObject; // for humans this is a weapon
+    [BitField<bool>(nameof(IsTransformChanged), 1)] // Enable this flag after setting the transform
+    [FieldOffset(0x38)] public byte ObjectFlags;
```
Haselnussbomber: Most of the xrefs access this field as qword, so I think we should use ulong as underlying type.

```suggestion
    [FieldOffset(0x38)] public ulong ObjectFlags;
```

FFXIVClientStructs/FFXIV/Client/Graphics/Scene/Object.cs
```diff
@@ -20,8 +23,23 @@ public SiblingEnumerator ChildObjects
     [FieldOffset(0x60)] public Quaternion Rotation;
     [FieldOffset(0x70)] public Vector3 Scale;
 
+    [VirtualFunction(0)]
+    public partial void Dtor(byte freeFlags);
```
Haselnussbomber: I *think* it returns `thisPtr`, even if it's not used in the function `DisableDraw` calls.
The `rax` register holds the return value and `mov rax, rbx` a couple instructions before `ret` moves `thisPtr` there. 🤔

Update: @wolfcomp said all Ctors and Dtors return this. It's what MSVC does.

```suggestion
    public partial Object* Dtor(byte freeFlags);
```

FFXIVClientStructs/FFXIV/Client/Graphics/Scene/Weapon.cs
```diff
@@ -25,12 +26,18 @@ public unsafe partial struct Weapon {
...
         [FieldOffset(0x02)] public byte Variant;
         [FieldOffset(0x03)] public byte Stain0;
         [FieldOffset(0x04)] public byte Stain1;
     }
+
+    [MemberFunction("E8 ?? ?? ?? ?? 48 89 07 4C 8B E0")]
+    public static partial Weapon* Create(WeaponModelId* weaponModelId);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 49 8B 16 48 8B 0D ?? ?? ?? ??")]
+    public partial byte Initialize(WeaponModelId* weaponModelId);
```
Haselnussbomber: It only returns 0 or 1, so it's most likely a bool. :)

```suggestion
    public partial bool Initialize(WeaponModelId* weaponModelId);
```

Haselnussbomber: I noticed it accesses a ~~bool~~ byte after the size of `WeaponModelId` (0x08) for the `Client::Graphics::Scene::CharacterBase_Initialize` call at 0x14043823B. 🤔
Maybe a wrapper struct carring an extra field?

Update: I found that this byte is set to AnimationVariant.

PR #1757, opened 2026-03-18
Move Ghidra scripts from Jython to PyGhidra

0x0ade:
Fedora doesn't even ship Python 2 packages anymore, which pushed me to the edge to figure out what's going wrong with PyGhidra, and from my testing:
- jpype picked the wrong overload on toAddr, so I simply wrapped it in a func that always explicitly casts to JLong first.
- FunctionDB.updateFunction was choking because jpype failed to turn Python lists into something that implements the Java List interface. Dunno if it's even meant to do that out of the box. Explicitly using ArrayList solves this.

### DIFF
--- a/ida/ffxiv_idarename.py
+++ b/ida/ffxiv_idarename.py
@@ -1,7 +1,7 @@
 # current exe version: 2020.12.29.0000.0000
 # @category __UserScripts
 # @menupath Tools.Scripts.ffxiv_idarename
-# @runtime Jython
+# @runtime PyGhidra
 
 from __future__ import print_function
 import os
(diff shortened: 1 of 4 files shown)


ida/ffxiv_idarename.py
```diff
@@ -330,6 +330,17 @@ def format_func_name(self, ea, current_func_name, proposed_func_name, class_name
...
+
+        # Selftest and hackfix for int / long overload conflicts in current year.
+        try:
+            toAddr(0x140000000)
+        except OverflowError:
+            import jpype.types
+
+            toAddrOrig = toAddr
+            def toAddr(ea):
+                return toAddrOrig(jpype.types.JLong(ea))
```
pohky:
```suggestion
        def toAddr(offset): # noqa
            # type: (int) -> Address
            return currentProgram.getAddressFactory().getDefaultAddressSpace().getAddress(offset) # noqa
```
i think just overriding it with the actual implementation here is better

pohky (approved):

0x0ade:
Rebased with your suggestion, (as per our chat on Discord) added a Windows example path, and expanded on the instructions a bit further. I didn't test this with a system Python PyGhidra install, but I hope it's sensible enough for anyone who does have such a setup to follow anyway.

PR #1764, opened 2026-03-27
SpawnNpcPacket: Fill out some more fields

redstrate:
Upstreaming the link-related stuff I figured out, and also found that the second u32 here is used for quests.


FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
```diff
@@ -9,13 +9,17 @@ namespace FFXIVClientStructs.FFXIV.Client.Game.Network;
 [StructLayout(LayoutKind.Explicit, Size = 0x290)]
 public partial struct SpawnNpcPacket {
     [FieldOffset(0x00)] public uint GimmickId;
-    [FieldOffset(0x04)] private uint Unk4; // Character+0x1BC0
+    /// <remarks> Instance ID of an Event NPC. Seen for Event NPCs that turn into Battle NPCs (during quests, for example.) </remarks>
```
redstrate, line 12: Not suuuper happy with this name, hence why I didn't add it to Character yet... having a field named `EventNpcId` there would be strange?

PR #1769, opened 2026-04-09
Add OpenContextMenuForMember to AgentFreeCompany

AtmoOmen:


FFXIVClientStructs/FFXIV/Client/UI/Agent/AgentFreeCompany.cs
```diff
@@ -28,6 +28,9 @@ public unsafe partial struct AgentFreeCompany {
 
     [FieldOffset(0xA90)] public FreeCompanyActionTimer ActionTimeRemaining;
 
+    [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 48 8D 4F 10 E8 ?? ?? ?? ?? 44 8B 43 20")]
+    public partial bool OpenContextMenuForMember(ushort memberIndex);
```
wolfcomp: Returns void, takes only byte in.
Missing entry in data.yml

wolfcomp (changes requested):
Also fields that can be documented inside the function.

PR #1773, opened 2026-04-12
AtkUnitBase.RegisterEvent

MidoriKami:
IDA seems to think there's a return value, but I have not found it be used anywhere.

Address was already mapped in data.yml

```c
Component::GUI::AtkResNode *__fastcall Component::GUI::AtkUnitBase_RegisterEvent(
        Component::GUI::AtkUnitBase *a1,
        Component::GUI::AtkEventType a2,
        unsigned int a3,
        Component::GUI::AtkEventListener *listener,
        Component::GUI::AtkResNode *nodeParam)
{
  Component::GUI::AtkResNode *result; // rax

  result = a1->RootNode;
  if ( result )
    return Component::GUI::AtkEventManager_RegisterEvent(
             &result->AtkEventManager,
             a2,
             a3,
             nodeParam,
             result,
             listener,
             0);
  return result;
}
```

### DIFF
--- a/FFXIVClientStructs/FFXIV/Component/GUI/AtkUnitBase.cs
+++ b/FFXIVClientStructs/FFXIV/Component/GUI/AtkUnitBase.cs
@@ -265,6 +265,9 @@ public unsafe partial struct AtkUnitBase : ICreatable {
     [MemberFunction("E8 ?? ?? ?? ?? 4D 8B CF C6 44 24 ?? ?? 4C 8B C7"), GenerateStringOverloads]
     public partial void SaveAddonConfig(CStringPointer name, bool a2, bool a3);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 8D 56 0C 48 8B CF")]
+    public partial void RegisterEvent(AtkEventType eventType, uint param, AtkEventListener* listener, AtkResNode* node);
+
     [VirtualFunction(3)]
     public partial bool Open(uint depthLayer);
 


FFXIVClientStructs/FFXIV/Component/GUI/AtkUnitBase.cs
```diff
@@ -265,6 +265,9 @@ public bool IsVisible {
     [MemberFunction("E8 ?? ?? ?? ?? 4D 8B CF C6 44 24 ?? ?? 4C 8B C7"), GenerateStringOverloads]
     public partial void SaveAddonConfig(CStringPointer name, bool a2, bool a3);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 8D 56 0C 48 8B CF")]
+    public partial void RegisterEvent(AtkEventType eventType, uint param, AtkEventListener* listener, AtkResNode* node);
```
wolfcomp, line 269: Does return `AtkEvent *` from inner call of `AtkEventManager.RegisterEvent`

wolfcomp (changes requested):

FFXIVClientStructs/FFXIV/Component/GUI/AtkUnitBase.cs
MidoriKami: Doesn't seem right to me for several reasons.

1. In the function the "return value"'s type isn't consistent.
2. Nowhere is the return value used from this function
3. There's no use or benefit of having a AtkEvent*, it gets attached to the root node.

I do not think this function should have a return value. @Haselnussbomber second opinion?

Haselnussbomber: Yeah, I'm not sure. The compiler may have optimized it this way, so that either a nullptr from the node check stays in rax, or the result of the inner RegisterEvent call.

The same happens for
- `Component::GUI::AtkTimer.RegisterEvent`
- `Component::GUI::AtkSimpleTween.RegisterEvent` and
- `Component::GUI::AtkAddonControl.RegisterEvent`.

I'd probably also set it to `void`, given that nothing makes use of it.

PR #1777, opened 2026-04-18
AtkComponentListItemPopulator Fix Incorrect Types

MidoriKami:
Through testing with AddonSelectString, it passes a PopupMenuDerive as the event listener, not an AtkUnitBase.


MidoriKami:
This is a known an intentional breaking change, the only known consumer of this code is KTK.

PR #1787, opened 2026-04-28
Update AtkStage

MidoriKami:


MidoriKami:
Confirmed the total size is increased by 0xA8
```c
  }
  Client::System::String::Utf8String_dtor((a1 + 152));
  *(a1 + 120) = &vtbl_Component::GUI::AtkEventListener;
  *a1 = vtbl_Component::GUI::AtkEventTarget;
  if ( (a2 & 1) != 0 )
    FreeMemory(a1, 0x75EA8ui64);
  return a1;
}
```

PR #1804, opened 2026-05-02
Fix mistyped CharacterBase->Materials

Exter-N:
Per 7.5's `14043C8C8` and `14043C8DF`, `CharacterBase->Materials` has to point to a subtype of `ResourceHandle`.


wolfcomp:
`14043C8C8` seems to reference `PerSlotStagingArea` and not `Materials`

Update: 
`14043C8C8` aka `ResourceHandle_IncRef` references `SlotStagingArea.field_18`
`14043C8DF` aka `ResourceHandle_DecRef` references `Materials`

Exter-N:
`14043C8DF` is not a DecRef call, it is *after* a DecRef call, and stores the pointer that was just IncRef'd into `Materials`

PR #1809, opened 2026-05-02
Waitingway struct updates

WorkingRobot:
Probably a breaking change due to QueueInfoState.AverageWaitTime, but it was mistakenly set to a boolean for some reason in the past.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
@@ -61,6 +61,9 @@ public unsafe partial struct GameMain {
     [MemberFunction("E8 ?? ?? ?? ?? 32 DB 0F B6 C3")]
     public static partial bool IsInIdleCam();
 
+    [MemberFunction("E8 ?? ?? ?? ?? 45 84 ED 75 57")]
+    private partial void StartTerritoryTransition(uint localPlayerEntityId, uint nextTerritoryTypeId, nint a4, nint a5, nint a6, ushort conditionId, nint a8, nint a9);
+
     [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 8B 44 24 70 48 8D 8D ?? ?? ?? ??")]
     public partial void QueueActiveFestivals(Festival festival1, Festival festival2, Festival festival3, Festival festival4, Festival festival5, Festival festival6, Festival festival7, Festival festival8); // Applies once the current "event" is done (GPose, Cutscene etc)
 
--- a/FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
@@ -61,9 +61,6 @@ public unsafe partial struct GameMain {
     [MemberFunction("E8 ?? ?? ?? ?? 32 DB 0F B6 C3")]
     public static partial bool IsInIdleCam();
 
-    [MemberFunction("E8 ?? ?? ?? ?? 45 84 ED 75 57")]
-    private partial void StartTerritoryTransition(uint localPlayerEntityId, uint nextTerritoryTypeId, nint a4, nint a5, nint a6, ushort conditionId, nint a8, nint a9);
-
     [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 8B 44 24 70 48 8D 8D ?? ?? ?? ??")]
     public partial void QueueActiveFestivals(Festival festival1, Festival festival2, Festival festival3, Festival festival4, Festival festival5, Festival festival6, Festival festival7, Festival festival8); // Applies once the current "event" is done (GPose, Cutscene etc)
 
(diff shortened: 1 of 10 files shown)


FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
```diff
@@ -61,6 +61,9 @@ public unsafe partial struct GameMain {
     [MemberFunction("E8 ?? ?? ?? ?? 32 DB 0F B6 C3")]
     public static partial bool IsInIdleCam();
 
+    [MemberFunction("E8 ?? ?? ?? ?? 45 84 ED 75 57")]
+    private partial void StartTerritoryTransition(uint localPlayerEntityId, uint nextTerritoryTypeId, nint a4, nint a5, nint a6, ushort conditionId, nint a8, nint a9);
```
Haselnussbomber: Can you use the IClientState.ZoneInit event in Dalamud instead of hooking this?

FFXIVClientStructs/FFXIV/Client/Game/UI/ContentsFinder.cs
```diff
@@ -82,8 +95,16 @@ public unsafe partial struct ContentsFinderQueueInfo {
...
+    // Sends packet to withdraw from queue
     [MemberFunction("40 53 48 83 EC ?? 48 8B D9 0F B6 49 ?? 8D 41")]
     public partial void CancelQueue();
+
+    // Called when packet is received to withdraw from queue
+    [MemberFunction("4C 8B DC 53 41 56 41 57 48 81 EC ?? ?? ?? ?? 48 8B 05 ?? ?? ?? ?? 48 33 C4 48 89 84 24 ?? ?? ?? ?? 4C 8B F1")]
+    public partial void OnQueueWithdrawn(uint logMessageId, ulong contentId);
+
+    [MemberFunction("48 89 5C 24 ?? 57 48 83 EC 20 80 79 55 00")]
+    public partial void OnQueuePop(ContentsFinderQueueState newState, uint contentId, nint a4, byte isInProgressParty, ContentsFinder.LootRule lootRule, ulong inProgressPartyStartTimestamp, nint a8, bool isUnrestrictedParty, byte isMinimalIL, byte isSilenceEcho, byte isExplorerMode, byte isLevelSync, byte isLimitedLeveling);
```
Haselnussbomber: Do you not want to make those bytes bools? They are set to the fields anyway.

FFXIVClientStructs/FFXIV/Client/UI/Agent/LobbyUIClient.cs
```diff
@@ -15,6 +15,10 @@ public unsafe partial struct LobbyUIClient {
     [FieldOffset(0x30)] public StdVector<LobbyDataCenterWorldEntry> CurrentDataCenterWorlds;
 
     [FieldOffset(0x48)] public LobbySubscriptionInfo* SubscriptionInfo;
+
+    // Technically part of LobbyRequestCallback
+    [VirtualFunction(4)]
+    public partial void ReportError(LobbyStatusCode* status);
```
Haselnussbomber: Please make a struct for `Application::Network::LobbyClient::LobbyRequestCallback` and add it there.

FFXIVClientStructs/FFXIV/Component/GUI/AtkUldManager.cs
```diff
@@ -49,6 +49,15 @@ public unsafe partial struct AtkUldManager {
     [MemberFunction("F6 81 ?? ?? ?? ?? ?? 44 8B CA 74 42")]
     public partial AtkResNode* SearchNodeById(uint id);
 
+    [MemberFunction("E8 ?? ?? ?? ?? 48 8B C8 3B DD")]
+    public partial AtkResNode* SearchNodeByIndex(uint idx);
```
Haselnussbomber: As Kami mentioned, getting nodes by index is dangerous with all the native tweaks people do.
Is there a specific reason for adding this instead of getting nodes by id?

Haselnussbomber (changes requested):

FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
WorkingRobot: image
Technically, this hook gets called slightly sooner, but yeah, I can just use that event instead. Is there a reason it shouldn't be included in CS, though?

FFXIVClientStructs/FFXIV/Client/Game/UI/ContentsFinder.cs
WorkingRobot: Forgot to do that, my bad.

FFXIVClientStructs/FFXIV/Component/GUI/AtkUldManager.cs
WorkingRobot: Not really, but I suppose it's better to remove it to not make it a footgun for developers.

FFXIVClientStructs/FFXIV/Client/Game/GameMain.cs
Haselnussbomber: Dalamud hooks UIModule.HandlePacket which is called at 0x140862D4E, further down the calling function.
Yes, StartTerritoryTransition is called a bit earlier, but adding this would require figuring out the parameters. For example you have a couple nint instead of uint parameters and there are also just 7, not 9, in total.

It's not *that* bad actually, since we already have the ZoneInitPacket, but still:
Screenshot

If Dalamuds event works for you, *please* just use that.

PR #1813, opened 2026-05-04
Add IsInside field to GraphicsConfig

Abyeon:
Added IsInside field which tracks if the player is inside a house.


Haselnussbomber:
Lmao, I'm so tired i wrote the commit message wrong. I renamed it to IsIndoor based on the path snippet it checks.

PR #1817, opened 2026-05-07
Fix return type of TextModuleInterface.FormatUtf8String

Haselnussbomber:
It's the return value of TextModuleInterface.FormatString.

Not sure exactly, since it looks like it's checked for -1 as int at `74 ?? 85 C0 74 ?? 83 F8 ?? 75 ?? 0F B6 D0`, but maybe that was their initialization value of the local variable.


Haselnussbomber:
Unless I searched incorrectly, nobody is using it on GitHub: https://github.com/search?q=FormatUtf8String+NOT+is%3Afork+language%3AC%23&type=code&l=C%23

PR #1827, opened 2026-05-12
Fix CreateRenderModel params

Haselnussbomber:
1 parameter too much.


Haselnussbomber:
Merging it, since it doesn't work like that and only existed for 10 days.

PR #1851, opened 2026-06-01
Correct EventObject PlayAnimation method

MnFeN:


Haselnussbomber:
Pushed the added fields as https://github.com/aers/FFXIVClientStructs/commit/b8e89baf85cb4ee29f1fd61087ceff9918590029 to the main branch and fixed parameter types of the PlayAnimation method.
Those are all 4 byte registers, so uint.
Also fixed the xmldoc warnings.

In regards to hooking it in a plugin, there should be no issues since the last parameter is entirely unused.

redstrate:
@MnFeN Are you happy with the documentation I added to main now, or is there something else we want to integrate from this PR?

MnFeN:
> @MnFeN Are you happy with the documentation I added to main now, or is there something else we want to integrate from this PR?

It looks nice. By the way, based on your observation, could you explain why the `sharedTimelineState` param in EObjAnimation and MapEffect both seem to have no actual influence on the visual effect? It was quite confusing to me.

redstrate:
> > @MnFeN Are you happy with the documentation I added to main now, or is there something else we want to integrate from this PR?
> 
> It looks nice. By the way, based on your observation, could you explain why the `sharedTimelineState` param in EObjAnimation and MapEffect both seem to have no actual influence on the visual effect? It was quite confusing to me.

That's still confusing for me too, sorry! My _theory_ is that it's some kind of arbitrary variable that's somewhat correlated to the animations played, because parts of the code check if a different `sharedTimelineState` is set before continuing.

But then you'll encounter cases like `sharedTimelineState` being used in the SpawnObject packet, and _that_ definitely correlates to which timeline indices are played! So maybe the answer is "sharedTimelineState directly correlates to which timelines are being played, except in some exceptional cases"??

I would love to see a better breakdown of this system because its strange that they would have two variables that mean similar but slightly different things. That would involve staring at what retail does for a while though :-D

PR #1861, opened 2026-06-10
Change WKSMission Score type from ushort to uint

LeontopodiumNivale14:
WAs marked with a // TODO originally, doesn't properly return right with the new cosmic missions that go above the ushort limit


LeontopodiumNivale14:
I honestly don't know how to fix this except for it's MEANT to be a uint, not a ushort

Haselnussbomber:
Since you opened the PR from your forks main branch, I can't push to it, because it's protected.
I've merged it manually and added a new field to avoid breaking changes. See https://github.com/aers/FFXIVClientStructs/commit/d3bbb3d39c7a62e87ae1f687cbf9971a6def688e

PR #1869, opened 2026-07-06
Add AtkServer.UICommandList and ImmediateContext.DoSetTargetCommand

sourpuh:


FFXIVClientStructs/FFXIV/Component/GUI/AtkUICommand.cs
```diff
@@ -0,0 +1,53 @@
...
+public enum AtkUICommandFormat : ushort {
+    Default = 0,
+    // 1 & 2 change texture counts which use extended variants of AtkUICommandDraw.
+    ClipMask = 3,
+}
+
+[StructLayout(LayoutKind.Explicit, Size = 0x60)]
+public unsafe struct AtkUICommandClipMask {
+    [FieldOffset(0x00)] public AtkUICommandType Type;
+    [FieldOffset(0x04)] public AtkUICommandFormat Format;
```
Haselnussbomber, line 35: Is Format on every command?
I suggest making a `AtkUICommand` struct and let the command structs inherit from it.
That way you can use a single `AtkUICommand*` field in `AtkUICommandEntry`.

See https://github.com/aers/FFXIVClientStructs/commit/25cf561c1a28586b21c5df54078047b9ada92a67 where I just did it for RenderCommand.

FFXIVClientStructs/FFXIV/Client/Graphics/Kernel/ImmediateContext.cs
```diff
@@ -84,6 +84,9 @@ public unsafe partial struct ImmediateContext {
     [MemberFunction("48 89 6C 24 ?? 56 48 83 EC 40 48 83 7A")] // inlined in some places
     public partial void DoClearCommand(RenderCommandClear* clearCommand);
 
+    [MemberFunction("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? B1 48")]
+    public partial void DoSetTargetCommand(RenderCommandSetTarget* command);
```
Haselnussbomber, line 88: Please add this to the data.yml.

FFXIVClientStructs/FFXIV/Component/GUI/AtkUICommand.cs
sourpuh: Yes, Format is on every command. Good suggestion; I wasn't aware it was possible to inherit. I assume you meant for me to drop the CExporterUnion fields.

PR #1881, opened 2026-07-14
Untangle visibility, targetable, and other fields in spawn packets

redstrate:
These various fields are really strangely defined and were bothering me for a while. So I took the time to finally untangle them and hopefully they make more sense now.

The first one is targetable status, this plugs into the GameObject field of the same name but is *not* the flag itself. The disassembly (from Ghidra) looks daunting but what it's actually doing is checking if the field is 0 - or false - and if so, setting two flags: Unk0 | IsTargetable. Higher values don't do anything, I confirmed by recreating the code in Godbolt.

The next field is Visibility, it also plugs into the GameObject field of the same name but the name is somewhat confusing. On first glance, you would think every object spawned would have a Visibility of 1 - or true - but the field means the opposite! If you spawn an object with Visibility set to 0, its shown and vice versa. I renamed this field to something more suitable. I didn't bother changing the GameObject field for now.

There's a couple of fields that were included in the struct but not read by the client, so I marked those to be removed at a later date.

The final two fields are in SpawnObjectPacket, and are meant for general purpose across all types of objects. For example, the initial SharedGroupTimeline state is passed in Arg1 for EventObj. I renamed those fields as their actual purpose is already known in a way.

The remaining changes is just helpers to make the functions these fields plug into more readable. A lot of the GameObject methods I stumbled upon use this strange update flag system, so I started detailing that but didn't put too much effort into that yet.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
@@ -122,9 +122,11 @@ public partial struct CommonSpawnData {
 [StructLayout(LayoutKind.Explicit, Size = 0x40)]
 public struct SpawnObjectPacket {
     [FieldOffset(0x00)] public byte ObjectIndex;
-    [FieldOffset(0x01)] public byte ObjectKind;
-    [FieldOffset(0x02)] public byte TargetableStatus;
-    [FieldOffset(0x03)] public byte Visibility;
+    [FieldOffset(0x01)] public byte ObjectKind; // TODO: Use ObjectKind enum
+    [FieldOffset(0x02), Obsolete("Use IsTargetable")] public byte TargetableStatus;
+    [FieldOffset(0x02)] public bool IsTargetable;
+    [FieldOffset(0x03), Obsolete("Use IsHidden")] public byte Visibility;
+    [FieldOffset(0x03)] public bool IsHidden;
     /// <remarks> <see cref="HousingObjectId"/> when <see cref="ObjectKind.HousingEventObject"/>. </remarks>
     [FieldOffset(0x04)] public uint BaseId;
     [FieldOffset(0x08)] public uint EntityId;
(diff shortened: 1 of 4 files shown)


FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
```diff
@@ -136,16 +138,20 @@ public struct SpawnObjectPacket {
     /// <remarks> Used for <see cref="ObjectKind.EventObj"/>. </remarks>
     [FieldOffset(0x18)] public uint GimmickId;
     [FieldOffset(0x1C)] public float Radius;
-    [FieldOffset(0x20)] private ushort Unk20; // SharedGroupTimelineState?
+    [FieldOffset(0x20)] private ushort Unk20; // TODO: Remove, this isn't read
```
Haselnussbomber: Removing private fields is fine. It's not a breaking change. Unk27 to Unk30 can be removed too.

FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
```diff
@@ -136,16 +138,20 @@ public struct SpawnObjectPacket {
...
-    [FieldOffset(0x2C)] private uint Unk2C;
-    [FieldOffset(0x30)] private uint Unk30;
+    [FieldOffset(0x27)] private byte Unk27; // TODO: Remove, this isn't read
+    [FieldOffset(0x28)] private uint Unk28; // TODO: Remove, this isn't read
+    [FieldOffset(0x2C), Obsolete("Use Arg1")] private uint Unk2C;
+    [FieldOffset(0x30), Obsolete("Use Arg2")] private uint Unk30;
+    /// <remarks> Used by <see cref="ObjectKind.GatheringPoint"/>, <see cref="ObjectKind.EventObj"/>, <see cref="ObjectKind.AreaObject"/>. </remarks>
+    [FieldOffset(0x2C)] private uint Arg1;
+    /// <remarks> Used by <see cref="ObjectKind.GatheringPoint"/>, <see cref="ObjectKind.EventObj"/>, <see cref="ObjectKind.HousingEventObject"/>. </remarks>
+    [FieldOffset(0x30)] private uint Arg2;
```
Haselnussbomber: Don't you want to make them public? If not, maybe name them UnkArg1, UnkArg2.

FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs
```diff
@@ -275,6 +280,18 @@ public unsafe partial struct GameObject {
     [MemberFunction("E8 ?? ?? ?? ?? 84 C0 74 ?? 33 FF 48 89 B4 24")]
     public partial bool IsSharedGroupLoaded();
 
+    /// <summary>Sets EventState to a new value.</summary>
+    [MemberFunction("80 89 ?? ?? ?? ?? ?? 88 51")]
+    public partial void SetEventState(byte eventState);
+
+    /// <summary>Sets Visibility to a new value.</summary>
+    [MemberFunction("E8 ?? ?? ?? ?? 8B 4F ?? 89 8E")]
+    public partial void SetVisibility(byte visibility);
```
Haselnussbomber, line 290: This is weird... Passing 1 hides the character, passing 0 shows it. 🤔
Sorry, I just read the PR message afterwards. 😅

redstrate: Yeah... I also ended up writing that function first - so I forgot to put that quirk in the documentation. Fixed now!

FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
redstrate: I do, same reason as before I didn't realize they were private. Fixed

FFXIVClientStructs/FFXIV/Client/Game/Network/SpawnPackets.cs
redstrate: Done, I blinded the `private` there :D

PR #1882, opened 2026-07-15
Fix casing in Addon attribute for SelectYesno

Knightmore:
This has been wrong the whole time, no?
@Haselnussbomber your Debugger is also picking it up now with that change.


FFXIVClientStructs/FFXIV/Client/UI/AddonSelectYesno.cs
```diff
@@ -5,7 +5,7 @@ namespace FFXIVClientStructs.FFXIV.Client.UI;
 // Client::UI::AddonSelectYesNo
```
wolfcomp: You should change this comment at the same time.

PR #1884, opened 2026-07-15
Add Atk collision stuff

Zeffuro:
Some collision stuff we found, a lot of this can be verified when looking at how the Action Menu works.

### DIFF
--- a/ida/data.yml
+++ b/ida/data.yml
@@ -2862,6 +2862,8 @@ classes:
       0x14066E700: GetDuplicatedNode
       0x14066AD80: UpdateDrawNodeList
       0x14066AFF0: TraverseNode
+      0x14066B060: CountDrawNodeCallback
+      0x14066B070: AddDrawNodeCallback
       0x14066D180: Finalizer
       0x140668920: SetupTimelineFromULDResourceHandle # CreateTimeline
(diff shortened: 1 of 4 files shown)


ida/data.yml
```diff
@@ -12973,6 +12979,8 @@ classes:
         base: Client::UI::Agent::AgentInterface
     funcs:
       0x140581210: ctor
+      0x1405812C0: Show
+      0x140581480: Hide
```
Haselnussbomber: No need for these. Those are vfuncs of AgentInterface.

ida/data.yml
```diff
@@ -21924,6 +21932,9 @@ classes:
         base: Component::GUI::AtkUnitBase
     funcs:
       0x14167C290: ctor
+      0x1416A3150: dtor
+      0x14167C3B0: ReceiveEvent
+      0x14167C750: OnRefresh
```
Haselnussbomber: Same here, not needed. These are vfuncs of AtkUnitBase.

Haselnussbomber (changes requested):

ida/data.yml
Zeffuro: Ah fair

Haselnussbomber (approved):

PR #1887, opened 2026-07-19
Change ZoneClient and ChatClient to public access

AtmoOmen:
given that the structures of ZoneClient and ChatClient have been largely reverse-engineered, making them public enables support for plugins such as those for latency testing.


pohky:
yeah but they are wrong so that needs to be fixed first

AtmoOmen:
> yeah but they are wrong so that needs to be fixed first

at least `Host`、`Port` in `ClientBase` look fine.
I could make another pr later, but I think it's better to make them public first. After all, they are not totally wrong.

PR #1893, opened 2026-07-22
Added camera objects and a new timeline var

Glorou:
just some camera stuff + a new timeline var


Haselnussbomber:
I've reverted the addition of the `TimelineRowIds` array because it looks like you overlooked the pointer math here.

`0xFC - 0x1C` = `0xE0`, the offset of the `TimelineIds` array.

Screenshot

Haselnussbomber:
Oh, actually I overlooked the dereference `*(_WORD *)TimelineRowIds`. Hm, I'll look into it. Sorry.

Haselnussbomber:
I don't know enough to name the arrays, but I've added them with commit e508667d.

PR #1899, opened 2026-07-26
Small bone constraint and pop range layout additions

redstrate:
**Disclosure**: this was prepared and submitted by a human agent acting on behalf of the account owner (who is the same person) the underlying analysis was performed with the owner's tooling and direction because its literally just 37 lines of changes. the owner also drank one cup of water and even plans to hydrate themselves later.

See individual commits, just random stuff while I'm in between things

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Graphics/Physics/ConstraintBase.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Graphics/Physics/ConstraintBase.cs
@@ -1,14 +1,28 @@
+using FFXIVClientStructs.FFXIV.Client.Graphics.Render;
+
 namespace FFXIVClientStructs.FFXIV.Client.Graphics.Physics;
 
 // Client::Graphics::Physics::ConstraintBase
 [GenerateInterop(isInherited: true)]
 [StructLayout(LayoutKind.Explicit, Size = 0x18)]
 public unsafe partial struct ConstraintBase {
+    [FieldOffset(0x08)] public Skeleton* Skeleton;
+
     /// <remarks> Set in constructor. </remarks>
     [FieldOffset(0x10)] public ConstraintType Type;
 
     [MemberFunction("E8 ?? ?? ?? ?? 48 8D 05 ?? ?? ?? ?? C7 43 ?? ?? ?? ?? ?? ?? ?? ?? 33 C0 48 89 43 ?? 48 89 43 ?? 89 43 ?? 48 8B C3 C7 43")]
     public partial ConstraintBase* Ctor(ConstraintType type);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 74 ?? F3 0F 10 44 24 ?? B0 ?? F3 0F 10 4C 24 ?? F3 0F 11 43")]
+    public partial bool SetSkeleton(Skeleton *skeleton);
+
+    [VirtualFunction(0)]
+    public partial ConstraintBase* Dtor(byte freeFlags);
+
+    /// <summary> Sets Skeleton back to null, decreases refcount if applicable. </summary>
+    [VirtualFunction(1)]
+    public partial void ResetSkeleton();
 }
 
 public enum ConstraintType : uint {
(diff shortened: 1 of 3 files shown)


ida/data.yml
```diff
@@ -26663,7 +26650,7 @@ classes:
       0: dtor
       1: ClearSkeleton
     funcs:
-      0x141931480: ctor
+      0x141931480: Ctor
```
redstrate: FWIW I changed it like this so the API doesn't break, although in the future do we prefer it written like "ctor" or "Ctor"? At least for the dtor vfuncs, I see "dtor" everywhere.

Haselnussbomber: Does this cause issues being different casing?
I intentionally lowercased all ctors and dtors in the data.yml and made CExporter export them lowercased, as well as lowercasing them in ResolverTester so that it can find the entries in the data.yml.

Related commits:
CExporter: 4d8d5be157931a7d963959afc171f69e8043af68
ResolverTester: f2e542007b23622b4926f516dc894be4919e6c25
ctors in data.yml: 66495d391b2e4742d76c462cffcbe824ee5029dd
dtors in data.yml: 1a98ba8332f756f415bf02f3c182e2f358774469

Haselnussbomber: Okay, funny I was commenting on this too. We should keep them lowercased.
The data.yml is just for keeping track of functions between exes and for the idarename script.
Those entries have nothing to do with the C# code, so this is not API breaking.
CExporter will lowercase ctor/dtor automatically.

redstrate: Ehh, this is where I'm not sure. I had trouble with Ghidra being able to map the parameters of the Ctor unless the casing matched. I guess that's more of an issue in the script then?

Haselnussbomber: Oh, maybe I have overlooked some places in CExporter. I'll fix it.
Edit: Oh yeah, I forgot MemberFunctions. My bad.

redstrate: Thanks! I'll revert this change

FFXIVClientStructs/FFXIV/Client/Graphics/Physics/ConstraintBase.cs
```diff
@@ -1,14 +1,28 @@
...
+    [FieldOffset(0x08)] public Skeleton* Skeleton;
+
     /// <remarks> Set in constructor. </remarks>
     [FieldOffset(0x10)] public ConstraintType Type;
 
     [MemberFunction("E8 ?? ?? ?? ?? 48 8D 05 ?? ?? ?? ?? C7 43 ?? ?? ?? ?? ?? ?? ?? ?? 33 C0 48 89 43 ?? 48 89 43 ?? 89 43 ?? 48 8B C3 C7 43")]
     public partial ConstraintBase* Ctor(ConstraintType type);
+
+    [MemberFunction("E8 ?? ?? ?? ?? 84 C0 74 ?? F3 0F 10 44 24 ?? B0 ?? F3 0F 10 4C 24 ?? F3 0F 11 43")]
+    public partial void SetSkeleton(Skeleton *skeleton);
```
Haselnussbomber: The signature resolves to a function that is used as if condition. It always returns `true`, from what I can see.

FFXIVClientStructs/FFXIV/Client/LayoutEngine/Layer/PopRangeLayoutInstance.cs
```diff
@@ -0,0 +1,18 @@
+using System.Numerics;
```
Haselnussbomber: ~~You probably want to use `Common::Math::Vector3`.~~

Scratch that. The size is different, `System.Numerics.Vector3` fits here.

redstrate: Done

FFXIVClientStructs/FFXIV/Client/Graphics/Physics/ConstraintBase.cs
redstrate: Indeed yeah, I missed this. Fixed now

PR #1906, opened 2026-08-01
PartyList and Alliance stuff.

Zeffuro:
The SpecialPet slot actually seems to be used when I summon Carbuncle on my SMN

ffxiv_dx11_2026-08-01_23-10-53


Zeffuro:
Found one mismatch, let me fix that real quick

Zeffuro:
ffxiv_dx11_2026-08-02_00-04-13

~~Should be all good now unless you find any issues~~
Found one more issue, not everything shifted as far as I thought

Zeffuro:
Should be good now
ffxiv_dx11_2026-08-02_02-18-29
ffxiv_dx11_2026-08-02_02-18-56

PR #1908, opened 2026-08-02
Add a bit more BonePhysicsModule stuff I care about

redstrate:
I finally found the global that appears to help paper over issues >60 FPS, so I added that boolean. I'm unsure who it is under, BonePhysicsModule makes the most sense I think.

Additionally, I added the Initialize member function as that's another place where the g_60FPS global is used for certain calculations.


Haselnussbomber:
I renamed the global variable `g_60FPS` to `g_BoneSimulatorFrameRate`, because even though we have a `g_` prefix, in C++ and C# (and I guess in even more languages) you can't start a variable name with a number.
I hope this is fine like that. If the name is incorrect, please tell and I'll push a change to main branch with your suggestion.
Merging it for now.

redstrate:
> I renamed the global variable `g_60FPS` to `g_BoneSimulatorFrameRate`, because even though we have a `g_` prefix, in C++ and C# (and I guess in even more languages) you can't start a variable name with a number. I hope this is fine like that. If the name is incorrect, please tell and I'll push a change to main branch with your suggestion. Merging it for now.

Yeah that's fine! That honestly didn't pass my mind while writing that

PR #1909, opened 2026-08-03
add achievement packet

Jaksuhn:
sorry I forgot I should probably not be shoving a billion things into one pr


Jaksuhn:
thanks for the sig fix, my bad

PR #1912, opened 2026-08-04
Fix AgentTelepotTownData

AtmoOmen:


AtmoOmen:
`Flags` should be uint
image

and I don't believe someone is using this field (the whole struct has been wrong for months after all) so I think it's OK.

Haselnussbomber:
Yes, Flags should be uint, but the value range is within byte, so lets just avoid a breaking change for now in case someone does use it.

PR #1920, opened 2026-08-11
More layout stuff

redstrate:
One mistake and two follow-ups! See individual commits.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/LayoutEngine/Layer/WaterRangeLayoutInstance.cs
+++ b/FFXIVClientStructs/FFXIV/Client/LayoutEngine/Layer/WaterRangeLayoutInstance.cs
@@ -7,4 +7,6 @@ namespace FFXIVClientStructs.FFXIV.Client.LayoutEngine.Layer;
 [GenerateInterop]
 [Inherits<TriggerBoxLayoutInstance>]
 [StructLayout(LayoutKind.Explicit, Size = 0x90)]
-public unsafe partial struct WaterRangeLayoutInstance;
+public unsafe partial struct WaterRangeLayoutInstance {
+    [FieldOffset(0x80)] private uint Unk80; // First bit enables swimming, second bit messes with the environment like in The Tempest
+}
--- a/FFXIVClientStructs/FFXIV/Client/LayoutEngine/Layer/PositionMarkerLayoutInstance.cs
+++ b/FFXIVClientStructs/FFXIV/Client/LayoutEngine/Layer/PositionMarkerLayoutInstance.cs
@@ -1,10 +1,11 @@
 namespace FFXIVClientStructs.FFXIV.Client.LayoutEngine.Layer;
 
 // Client::LayoutEngine::Layer::PositionMarkerLayoutInstance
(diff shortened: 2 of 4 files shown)


FFXIVClientStructs/FFXIV/Client/LayoutEngine/Layer/WaterRangeLayoutInstance.cs
```diff
@@ -7,4 +7,6 @@ namespace FFXIVClientStructs.FFXIV.Client.LayoutEngine.Layer;
 [GenerateInterop]
 [Inherits<TriggerBoxLayoutInstance>]
 [StructLayout(LayoutKind.Explicit, Size = 0x90)]
-public unsafe partial struct WaterRangeLayoutInstance;
+public unsafe partial struct WaterRangeLayoutInstance {
+    [FieldOffset(0x80)] private uint Unk80; // First bit enables swimming, second bit enables... something?
```
redstrate: @pmgr Yeah so I played around with creating WaterRanges and they do what they say on the tin. I can corroborate that indeed the first bit is what controls everything underwater (VFX and swimming) but no idea about the second bit.

I've flopped and flipped it and yet couldn't figure out what it does? I didn't look very far tonight but all the instances in LGBs I seen only sets the first bit too.

pmgr: If I have it right, it uses those bits in separate, and combined (bit0 | bit1); the combination ties into shader keys, namely `GetUnderWaterLighting_Disable` and `GetUnderWaterLighting_Enable`, through EnvManager.

Essentially: `if (bit0 | bit1) GetUnderWaterLighting_Enable else GetUnderWaterLighting_Disable`

bit1 alone though, is also propagated separately into EnvManager, then gets consumed in `0x1405FCE30` and `0x141940900` (these are 7.55 funcs, I don't have latest bin analysed), the offset on EnvManager of interest is +0xBDC, which seems to be effectively affecting if a magnitude is set to 0.0 when bit1 is active, otherwise it's scaled by 0.1? (that 0.1 is stored in EnvManager+0xBD8)

So I'd say, maybe try setting that 0.1 to a high value, and then toggle the bit directly in EnvManager, see if you can repro some form of wind physics effect thing, because might be what this is, some form of wind physics toggle for water-fuckery? 

If you can't find anything, I guess just leaving it unlabeled is fine, it seems clearly consumed to me, but "unknown" is good enough lol

redstrate: Ah yeah okay it controlling some VFX/environment makes sense, I just checked the Tempest and there's two huge WaterRanges covering it with that bit set. My test zone is Limsa and it probably isn't the best place for testing that and it makes the sky murky :D

redstrate: Wrote that in a comment, having a vague idea what it controls is good enough for me ATM

PR #1926, opened 2026-08-19
ghidra: Add hack around getDataType("/float") failing

redstrate:
Testing against Ghidra 12.1.3 (and 12.1.2), getDataType("/float") doesn't work - but did work previously at some point. I don't know why it broke, and they didn't list this as an API change in their changelog.

While the program's datatype database doesn't work, it will work if we pull from the built-in one.

### DIFF
--- a/ida/ffxiv_structimporter.py
+++ b/ida/ffxiv_structimporter.py
@@ -1752,6 +1752,9 @@ if api is None:
 
                 dtm = currentProgram.getDataTypeManager()
                 dt = dtm.getDataType("/" + "/".join(syms))
+                # HACK: In Ghidra 12.1.3 getDataType("/float") fails, but it works if using BuiltInDataTypeManager
+                if dt is None:
+                    dt = BuiltInDataTypeManager.getDataTypeManager().getDataType("/" + "/".join(syms))
                 for i in range(pointer_count):
                     dt = dtm.getPointer(dt)
                 return dt


redstrate:
I hate this hack, but I couldn't live without having floats in the structs. Maybe I'm the only one hitting this issue? It's not even script related, the call will fail after just running the auto-analyze.

pohky:
i wasn't able replicate that issue (on both 12.1.2 and .3) with a fresh import but if more people are having the same issue we can merge this as a temp fix

redstrate:
> i wasn't able replicate that issue (on both 12.1.2 and .3) with a fresh import but if more people are having the same issue we can merge this as a temp fix

Can I ask how you're running Ghidra? Mine is kind of a homebrew mess, so it might be self-induced but I don't know

pohky:
> Can I ask how you're running Ghidra? Mine is kind of a homebrew mess, so it might be self-induced but I don't know

just running the latest release as-is, all the default analyzers that are pre-selected (except for the x86 constant reference one cause that just creates some issues with the rename script detecting vtable lengths)

redstrate:
> > Can I ask how you're running Ghidra? Mine is kind of a homebrew mess, so it might be self-induced but I don't know
> 
> just running the latest release as-is, all the default analyzers that are pre-selected (except for the x86 constant reference one cause that just creates some issues with the rename script detecting vtable lengths)

Can confirm with their release, I even left the auto analyzer settings to default and now I have a float datatype again! I'll fix this issue on my end.

(Also thanks for the tip about the x86 option!)

PR #1929, opened 2026-08-20
Add remaining interesting bits for the SupportDesk HTTP API

redstrate:
This will enable us to finally introspect the data for these requests! The most important bit was figuring out the stages and then making up a sensible delegate to hook into.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/SupportDesk/HttpConnection.cs
+++ b/FFXIVClientStructs/FFXIV/Client/SupportDesk/HttpConnection.cs
@@ -7,11 +7,20 @@ public unsafe partial struct HttpConnection {
     [VirtualFunction(0)]
     public partial HttpRequest* Dtor(byte freeFlags);
 
-    // Initializes the HTTP session.
+    /// <summary> Initializes the HTTP session. </summary>
     [VirtualFunction(1)]
     public partial void OpenSession();
 
-    // Cleans up the HTTP session.
+    /// <summary> Cleans up the HTTP session. </summary>
     [VirtualFunction(2)]
     public partial void CloseSession();
+
+    /// <summary> Creates a new HTTPRequest. </summary>
+    /// <param name="url">URL to make a request to.</param>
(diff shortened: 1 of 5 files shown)


FFXIVClientStructs/FFXIV/Client/SupportDesk/HttpConnection.cs
```diff
@@ -7,11 +7,20 @@ public unsafe partial struct HttpConnection {
...
     public partial void CloseSession();
+
+    /// <summary> Creates a new HTTPRequest. </summary>
+    /// <param name="url">URL to make a request to.</param>
+    /// <param name="verb">The HTTP verb, such as "GET" or "POST".</param>
+    /// <param name="data">Any request data to send (only relevant for POST requests really).</param>
+    /// <param name="dataLength">Length of the data buffer.</param>
+    /// <returns>The new HTTPRequest handle if it could be successfully created.</returns>
+    [VirtualFunction(4)]
+    public partial HttpRequest* MakeRequest(byte* url, byte* verb, byte* data, ulong dataLength);
```
redstrate: I'm a little unsure about this API, because I'm not really sure what kind of string it is. On the plugin side, I can do `Marshal.PtrToStringAnsi` on this and get a valid string, so is it a CStringPointer? And then `data` is technically always a valid string too as far as I've seen, maybe I should do the same there?

Haselnussbomber: `url` and `verb` should be CStringPointer, yeah. Make sure to add `[GenerateStringOverloads]`.
Personally, I think the `data` parameter should stay `byte*` as it has an extra `dataLength` parameter and is memcpy'd, so it might not be null terminated (which is a requirement for CStringPointer).

The same for BeginRequest in HttpRequestImpl where it gets passed to.

redstrate: Yup totally agree, changing now

PR #1938, opened 2026-08-28
Add classes for GimmickEventHandler and its children

redstrate:
This is a commonly-seen handler in instanced content, and handles a few director events (apart from the actual content director.) The one thing I was focused on is the 0x80000000 event - which apparently "activates" client-side GimmickRects like dungeon entrances.

Along the way I added missing classes under the Gimmick namespace, and a
few other things.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Event/EventHandler.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Event/EventHandler.cs
@@ -87,6 +87,14 @@ public unsafe partial struct EventHandler {
     [VirtualFunction(209)]
     public partial uint GetNameplateIconForObject(GameObject* gameObject);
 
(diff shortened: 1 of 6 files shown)


FFXIVClientStructs/FFXIV/Client/Game/Event/EventHandler.cs
```diff
@@ -87,6 +87,14 @@ public unsafe partial struct EventHandler {
     [VirtualFunction(209)]
     public partial uint GetNameplateIconForObject(GameObject* gameObject);
 
+    /// <summary> Some indicator of whether this EventHandler is currently active. For example, an inactive dungeon entrance GimmickRect doesn't do anything when walked into. </summary>
+    [VirtualFunction(213)]
+    public partial bool IsActive(void* a2);
```
pohky: a2 is the GameObject (called from vf38)

redstrate: Thanks! Fixed now

PR #1940, opened 2026-08-29
Add LayerId field to LayerManager

redstrate:
This is used to store the ID from the layer header.


redstrate:
Oh Ghidra didn't add the field from IManagerBase for some reason :facepalm:

PR #1942, opened 2026-08-31
fix preprocessing return type annotation

caitlyn-gg:
Fixes a return type annotation for Python < 3.14

### DIFF
--- a/ida/ffxiv_structimporter.py
+++ b/ida/ffxiv_structimporter.py
@@ -1242,7 +1242,7 @@ if api is None:
                     def is_virtual(self) -> bool:
                         return self.struct.virtual_functions is not None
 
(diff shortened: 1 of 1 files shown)


ida/ffxiv_structimporter.py
```diff
@@ -1242,7 +1242,7 @@ def __init__(self, struct: DefinedStruct):
                     def is_virtual(self) -> bool:
                         return self.struct.virtual_functions is not None
 
-                    def primary_base(self) -> 'BaseEdge' | None:
+                    def primary_base(self) -> 'BaseEdge | None':
```
wolfcomp:
```suggestion
                    def primary_base(self) -> BaseEdge | None:
```

`BaseEdge` doesn't need the apostrophe marks at all as it is defined before the class using the type as an anotation.

PR #1944, opened 2026-08-31
Adds terrain disk, scene, and render structures

universalconquistador:
Everything I need to support hit-testing in my plugin... and then some!

Shout out to winter and Lumina for the help.

NOTE: This PR also includes #1943. If that PR is not accepted, I'll revert that commit from this branch and remove the `TerrainResourceHandle.GetTeraFileData()` vfunc declaration.


universalconquistador:
I have updated this to keep the base `ResourceHandle` version of `GetData` marked as `[Obsolete]`, and also provide a specially-typed accessor that takes precedence in the IDA vtable.
image

PR #1947, opened 2026-09-09
Add BeastmasterGauge

wangel:
Adds `BeastmasterGauge` (ClassJobId 43).

Mapped at runtime on 7.56 with a dalamud plugin I wrote  that dumps every frame and
tracks per-byte min/max/distinct, then confirmed each field against the HUD or the combat log.

| Offset | Field | Notes |
|---|---|---|
| `0x08` | `TPGauge` | 0-250. No passive regen; granted by weaponskill combo bonuses. Instinctual skills spend the whole pool. |
| `0x09` | `FamiliarTPGauge` | 0-250. One pool shared across Battlehorn slots, not per-familiar. Arrives in +12 steps from the familiar's auto-attacks. |
| `0x0A` | `FamiliarTPAtLastUse` | Familiar TP immediately before the last familiar action, which scales its potency. |
| `0x0B` | `ActiveBattlehorn` | 0-3. The slot, not the creature. |
| `0x0C` | `InstinctualComboState` | Affinity, then `7` while Wavering Heart (Status 4643) locks out further combo advancement. |
| `0x0D` | `CurrentAffinity` | `BeastmasterAffinity`, 0-6. Clears ~7s after it is applied. |
| `0x0E` | `ChainCount` | Chained instinctual skills, matching the number drawn under the Inner Compass. |
| `0x0F` | `KinshipState` | Two nibbles: kin type high, Battlehorn slot low. |
| `0x10` | `InstinctState` | Two 2-bit counters: Natural Instinct low, Mastered Instinct high. Trait-gated, reads 0 below level 28. |

`BeastmasterAffinity` has six values the four compass affinities plus the two combo results:

```
None 0  Volant 1  Rampant 2  Durant 3  Eldritch 4  Sunstrider 5  Moonstalker 6
```

The compass runs clockwise Volant -> Rampant -> Durant -> Eldritch -> Volant, and a clockwise  pair inside the 7s window is what makes a combo. 

Sunstrider and Moonstalker are the results are the affinity you started from.  Odd = Sunstrider, Even = Moon.

The Status sheet backs this up. The six statuses the job applies are in this order (4595-4600). 

The upgrade only lands if the target survives the combo.

`0x10` holds two resources in one byte. Mastered Instinct (bits 2-3, trait Wild Heart III at
level 28) is gained when a you complete a combo...ie; You use tricks and then an axe skill.  Rally is the spender. 

Natural Instinct (bits 0-1, Wild Heart IV at level 40) is gained when a pet completes a combo....ie; you use an axe skill, then tricks. It's spent using Rallying Cheer. 

Each caps at 3, each spender clears only its own bits.

Thanks to @Taurenkey for identifying the `0x0C` `7` as Wavering Heart, and to @Haselnussbomber
for `XBMManager` and `ActionManager.BeastmasterPets`, which is where the familiar's identity
lives - the gauge only exposes the slot.

### DIFF
--- a/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
@@ -130,17 +130,7 @@ public enum SerpentCombo : byte {
     FourthLegacy = 6,
 }
 
-// The Inner Compass runs clockwise Volant -> Rampant -> Durant -> Eldritch -> Volant, so an
-// intentional combo is simply the next affinity in sequence, wrapping Eldritch back to Volant.
-// Completing one grants a fifth or sixth affinity, which the game calls the "two additional
-// instinctual affinities". Which of the two you get is the parity of the affinity the combo
-// started from: odd -> Sunstrider, even -> Moonstalker. Pairing those two in turn triggers
-// what the game calls an "infinitive combo".
-// The upgrade only lands if the target survives the combo. Kill it with the second skill and
-// you still get Wavering Heart and a ChainCount increment, but no Sunstrider or Moonstalker.
-// These values match the player statuses the job applies, which are contiguous and in this
-// same order: Status rows 4595 Volant Heart, 4596 Rampant Heart, 4597 Durant Heart,
-// 4598 Eldritch Heart, 4599 Sunstrider, 4600 Moonstalker.
+
 public enum BeastmasterAffinity : byte {
     None = 0,
     Volant = 1,
@@ -151,10 +141,7 @@ public enum BeastmasterAffinity : byte {
     Moonstalker = 6,
 }
 
-// A familiar's kin type, which decides which Kinship status Borrow grants and therefore which
-// ability Beast Mode turns into. Matches the order of the eight contiguous Kinship statuses,
-// Status rows 4602 Beast Kinship through 4609 Ash Kinship. All eight values observed directly
-// in the gauge, each cross-checked against the Kinship status the familiar actually granted.
+
 public enum BeastmasterKinType : byte {
     None = 0,
     Beastkin = 1,
--- a/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
@@ -129,3 +129,40 @@ public enum SerpentCombo : byte {
     ThirdLegacy = 5,
     FourthLegacy = 6,
 }
+
+// The Inner Compass runs clockwise Volant -> Rampant -> Durant -> Eldritch -> Volant, so an
+// intentional combo is simply the next affinity in sequence, wrapping Eldritch back to Volant.
+// Completing one grants a fifth or sixth affinity, which the game calls the "two additional
+// instinctual affinities". Which of the two you get is the parity of the affinity the combo
+// started from: odd -> Sunstrider, even -> Moonstalker. Pairing those two in turn triggers
+// what the game calls an "infinitive combo".
+// The upgrade only lands if the target survives the combo. Kill it with the second skill and
+// you still get Wavering Heart and a ChainCount increment, but no Sunstrider or Moonstalker.
+// These values match the player statuses the job applies, which are contiguous and in this
+// same order: Status rows 4595 Volant Heart, 4596 Rampant Heart, 4597 Durant Heart,
+// 4598 Eldritch Heart, 4599 Sunstrider, 4600 Moonstalker.
+public enum BeastmasterAffinity : byte {
+    None = 0,
+    Volant = 1,
+    Rampant = 2,
+    Durant = 3,
+    Eldritch = 4,
+    Sunstrider = 5,
+    Moonstalker = 6,
+}
+
+// A familiar's kin type, which decides which Kinship status Borrow grants and therefore which
+// ability Beast Mode turns into. Matches the order of the eight contiguous Kinship statuses,
+// Status rows 4602 Beast Kinship through 4609 Ash Kinship. All eight values observed directly
+// in the gauge, each cross-checked against the Kinship status the familiar actually granted.
+public enum BeastmasterKinType : byte {
+    None = 0,
+    Beastkin = 1,
+    Vilekin = 2,
+    Cloudkin = 3,
+    Seedkin = 4,
+    Wavekin = 5,
+    Scalekin = 6,
+    Soulkin = 7,
+    Ashkin = 8,
+}
--- a/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
@@ -198,6 +198,55 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
+// Beastmaster is a limited job; its gauge drives the Inner Compass UI.
+// Verified at runtime on patch 7.56 by observing JobGaugeManager while playing.
+[StructLayout(LayoutKind.Explicit, Size = 0x10)]
+public struct BeastmasterGauge {
+    // Beastmaster's own TP, 0-250. This does NOT regenerate passively - it is granted by the
+    // combo bonuses on the player's weaponskill chain (Axeblade Bite grants +13, Shieldsplitter
+    // +15), and a broken combo grants nothing. Instinctual skills spend the entire pool.
+    [FieldOffset(0x08)] public byte TPGauge;
+    // Familiar TP, 0-250. A single pool shared across Battlehorn slots - it is NOT reset by
+    // swapping familiars. Arrives in fixed +12 increments at irregular intervals (observed
+    // 1.3s, 3.1s, 19.2s and 3.0s apart), so it is not a server tick either; it tracks the
+    // familiar's auto-attacks. Familiar actions spend the entire pool.
+    [FieldOffset(0x09)] public byte FamiliarTPGauge;
+    // Familiar TP immediately before the most recent familiar action, which scales that
+    // action's potency. A familiar action expends the ENTIRE familiar pool - FamiliarTPGauge
+    // drops to 0 on the same frame - so "TP consumed" and "TP at time of use" are the same
+    // number and this field is both. Seen taking 132 and 168 on actions that zeroed a gauge
+    // holding exactly 132 and 168.
+    [FieldOffset(0x0A)] public byte FamiliarTPAtLastUse;
+    // Which Battlehorn slot's familiar is summoned; 0 when none. This is the SLOT, not the
+    // creature - the summoned beast's identity is not exposed here.
+    [FieldOffset(0x0B)] public byte ActiveBattlehorn;
+    // Reads the active affinity, then 7 while an instinctual combo resolves, then - if the
+    // pair was clockwise and so formed an intentional combo - the resulting Sunstrider or
+    // Moonstalker affinity ~2.2s later. If the pair was not clockwise it returns to 0 from 7
+    // instead. The 7 is not an affinity: it marks the "Wavering Heart" status (Status row
+    // 4643, "Otherwise engaged. Unable to perform combos with your familiar."), which locks
+    // out further combo advancement until it clears. That is why this stays a byte while
+    // CurrentAffinity below is typed as the enum.
+    [FieldOffset(0x0C)] public byte InstinctualComboState;
+    // Affinity of the most recent instinctual skill, replaced by the resulting Sunstrider or
+    // Moonstalker affinity once an intentional combo resolves. Cleared ~7s after the last
+    // skill. Only ever observed holding 0-6.
+    [FieldOffset(0x0D)] public BeastmasterAffinity CurrentAffinity;
+    // Number of instinctual skills chained so far, shown beneath the Inner Compass.
+    // Increments on any instinctual combo, not only intentional (clockwise) ones, and
+    // increases the extra damage dealt.
+    [FieldOffset(0x0E)] public byte ChainCount;
+    // Identifies the familiar whose Kinship is currently active: high nibble is its kin type,
+    // low nibble is the Battlehorn slot it was summoned from. Set when Borrow grants a Kinship
+    // and cleared to 0 when that Kinship is lost. Latched at the moment Borrow is used, so the
+    // low nibble is NOT a mirror of ActiveBattlehorn - swapping familiars afterwards moves
+    // ActiveBattlehorn while this stays put.
+    [FieldOffset(0x0F)] public byte KinshipState;
+
+    public BeastmasterKinType KinshipKinType => (BeastmasterKinType)(KinshipState >> 4);
+    public byte KinshipBattlehorn => (byte)(KinshipState & 0x0F);
+}
+
 #endregion
 
 #region Tanks
--- a/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
+++ b/FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
@@ -198,72 +198,23 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
-// Beastmaster is a limited job; its gauge drives the Inner Compass UI.
-// Verified at runtime on patch 7.56 by observing JobGaugeManager while playing.
+[GenerateInterop]
 [StructLayout(LayoutKind.Explicit, Size = 0x18)]
-public struct BeastmasterGauge {
-    // Beastmaster's own TP, 0-250. This does NOT regenerate passively - it is granted by the
-    // combo bonuses on the player's weaponskill chain (Axeblade Bite grants +13, Shieldsplitter
-    // +15), and a broken combo grants nothing. Instinctual skills spend the entire pool.
+public partial struct BeastmasterGauge {
     [FieldOffset(0x08)] public byte TPGauge;
-    // Familiar TP, 0-250. A single pool shared across Battlehorn slots - it is NOT reset by
-    // swapping familiars. Arrives in fixed +12 increments at irregular intervals (observed
-    // 1.3s, 3.1s, 19.2s and 3.0s apart), so it is not a server tick either; it tracks the
-    // familiar's auto-attacks. Familiar actions spend the entire pool.
     [FieldOffset(0x09)] public byte FamiliarTPGauge;
-    // Familiar TP immediately before the most recent familiar action, which scales that
-    // action's potency. A familiar action expends the ENTIRE familiar pool - FamiliarTPGauge
-    // drops to 0 on the same frame - so "TP consumed" and "TP at time of use" are the same
-    // number and this field is both. Seen taking 132 and 168 on actions that zeroed a gauge
-    // holding exactly 132 and 168.
     [FieldOffset(0x0A)] public byte FamiliarTPAtLastUse;
-    // Which Battlehorn slot's familiar is summoned; 0 when none. This is the SLOT, not the
-    // creature - the summoned beast's identity is not exposed here.
-    [FieldOffset(0x0B)] public byte ActiveBattlehorn;
-    // Reads the active affinity, then 7 while an instinctual combo resolves, then - if the
-    // pair was clockwise and so formed an intentional combo - the resulting Sunstrider or
-    // Moonstalker affinity ~2.2s later. If the pair was not clockwise it returns to 0 from 7
-    // instead. The 7 is not an affinity: it marks the "Wavering Heart" status (Status row
-    // 4643, "Otherwise engaged. Unable to perform combos with your familiar."), which locks
-    // out further combo advancement until it clears. That is why this stays a byte while
-    // CurrentAffinity below is typed as the enum.
+    [FieldOffset(0x0B)] public byte ActiveBattlehornIndex;
     [FieldOffset(0x0C)] public byte InstinctualComboState;
-    // Affinity of the most recent instinctual skill, replaced by the resulting Sunstrider or
-    // Moonstalker affinity once an intentional combo resolves. Cleared ~7s after the last
-    // skill. Only ever observed holding 0-6.
     [FieldOffset(0x0D)] public BeastmasterAffinity CurrentAffinity;
-    // Number of instinctual skills chained so far, shown beneath the Inner Compass.
-    // Increments on any instinctual combo, not only intentional (clockwise) ones, and
-    // increases the extra damage dealt.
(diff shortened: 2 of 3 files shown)


FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,38 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
+// Beastmaster is a limited job; its gauge drives the Inner Compass UI.
+// Verified at runtime on patch 7.56 by observing JobGaugeManager while playing.
+[StructLayout(LayoutKind.Explicit, Size = 0x10)]
+public struct BeastmasterGauge {
+    // Beastmaster's own TP. Weaponskills expend all of it; regenerates ~14 per tick.
```
wolfcomp: This doesn't correctly say how it is affected as it is from `Axeblade Bite` and `Shieldsplitter` that you get the respective `13` and `15` TP from each time it is executed.

Taurenkey:
The part you're not sure about should relate to the status you get as a player, when it's set to 7 you have the "Wavering Heart" status which prevents further advancement to the combo until it's cleared. Most likely to prevent triple weaving shenanigans breaking the gauge or something.

wangel:
That's it exactly, thank you!!!!

 "Otherwise engaged. Unable to perform combos with your familiar."

It also explains something that had been nagging me. I'd measured that hold at 2.22–2.23s and got the same number on every combo across three different pairings, which felt far too clean for a resolution delay. It isn't one, it's a fixed status duration.

Chasing it turned up a related bit of behaviour worth recording: if the second skill *kills* the target, you still get Wavering Heart and the `ChainCount` increment, but the affinity never upgrades to Sunstrider/Moonstalker. That had me convinced I had something broken.

Pointing me at the status side was worth doing twice over, because going back to the sheet turned up two more things.

First, the six affinities are contiguous and in the same order as the gauge values (4595 Volant Heart → 4600 Moonstalker), which independently confirms `Rampant = 2` — the one value I had gotten  from  the compass position rather than actually observed. Sunstrider and Moonstalker also each say that using a skill with the *other's* affinity triggers an "infinitive combo", so there's a tier above intentional that pairs 5 against 6.

Second, and bigger: I'd written `0x0F` off as a timer that goes up. It isn't a timer at all — I was reading it as a decimal. In hex it's two nibbles, high is the familiar's kin type and low is the Battlehorn slot it came from. I summoned one familiar of each kin and used Borrow:

```
0x11 Beast/1   0x12 Beast/2   0x83 Ash/3     0x51 Wave/1
0x32 Cloud/2   0x73 Soul/3    0x62 Scale/2   0x41 Seed/1
0x21 Vile/1
```

All eight kin types, each matching the familiar I had out and the Kinship status I actually got, numbered the same as the Kinship block at 4602–4609. So the branch now carries `KinshipState` and a `BeastmasterKinType` enum — kin type is in the gauge after all, which the original body said it wasn't.

While I was at it I also had to correct both TP fields, since neither gauge regenerates: player TP comes from the weaponskill combo bonuses (Axeblade Bite +13, Shieldsplitter +15), which averages neatly to the "~14 a tick" I had wrongly written down.

Branch and body are both updated.

wangel:
Two updates and a thank-you.

**`Rampant = 2` is no longer inferred.** I hooked `UseAction` in my dump plugin and attributed
each `0x0D` transition to the press just before it. Avalanche Axe applies `2` directly, so every
value in the enum has now been seen in the gauge rather than derived from Status sheet ordering.
I've corrected that line in the description above.

The same attribution mapped the four standalone axes, one per compass affinity:

| action | affinity |
|---|---|
| 44884 Avalanche Axe | Rampant (2) |
| 44887 Mistral Axe | Durant (3) |
| 44888 Spinning Axe | Eldritch (4) |
| 44889 Gale Axe | Volant (1) |

These are fixed per action — the affinity is shown on the actions in game, and only the
familiar's varies. Nothing this PR needs, but it is independent confirmation that the enum
ordering is real rather than a coincidence of the Status sheet: the four axes cover all four
affinities, so there is always one at `beast affinity + 1` and an intentional combo is available
whatever familiar is out.

The familiar's side resolves through the `XBMPet` sheet rather than the creature: affinity is a
property of the pet's **`AutoAttack` action** (see the sheet columns below), not of the pet
or its kin. Cu Sith and Puk share
`AutoAttack` `45187` and both apply Rampant despite differing in kin, `Pet` id and notebook
slot. Roughly 15 action ids cover all 50 pets. The unnamed `0x1A` is *not* the affinity - Cu Sith
is 3 and Puk is 2, and both give Rampant.

**Thanks @Haselnussbomber for `XBMManager` (4aa267bf)** — that joined up two things I had been
poking at separately. Some runtime detail back, in case it is useful to whoever writes the
binding for `ActionManager::GetBeastmasterPet` (`0x1408F0590`, named in the same commit):

- the argument is a Battlehorn slot, 0-based — 0, 1, 2. (Correction to an earlier draft of this:
  it is bounds-checked and returns 0 above index 2. I had called it through a 64-bit signature and
  was reading stale register bits above the returned byte.)
- the return is the **Monster Notebook position, 1-based**: 1 for a familiar sitting at notebook
  slot 1, 20 for one at slot 20.
- those values are the id `IsPetUnlocked` takes, and the bit index is `petId - 1`.
- `BNpcBase BaseId = petId + 18915` held on all five familiars I tried, across petId 1–20.
  `PetBuddy.DataID` does *not* track it consistently — one familiar is off by one, so `BaseId`
  is the field that stays in step.

On `XBMPet` (51 rows, row 0 null): I had worked out some columns by matching the sheet against
familiars I had summoned, but nt153133 kindly sent me the SaintCoinach definition, which names
them properly and corrects two of my guesses. In dump order:

`Description`, `TrickDescription`, `TemperedReleaseDescription`, `Icon`, `Pet`, `AutoAttack`,
`NaturalHabitat`, `Classification`, two unnamed bytes at `0x19` and `0x1A`, `NaturalHabitatType`.

`Pet` is the link to the `Pet` sheet - that is the id `IBuddyList.PetBuddy.DataID` gives you.
`AutoAttack` is an `Action` row. `NaturalHabitat` is a complexlink: `PlaceName` when
`NaturalHabitatType` is 1, `ContentFinderCondition` when it is 2. What I had taken for a tier was
the habitat type, and what I had taken for a text reference was the icon.

`Classification` is the kin type, 1-8 Beast/Vile/Cloud/Seed/Wave/Scale/Soul/Ash - that one I had
right, and it matches the gauge's `0x0F` high nibble exactly.

One correction I guessed `XBMModule` was where the
slot→creature mapping lived. `GetBeastmasterPet` is the actual answer.

wangel:
Rewrote the description in my own words - the original was written by AI was slop, I was being lazy with it.  The data was all gathered from a plugin I wrote, that I used while testing and playing.  If you any of you want the plugin to see it, hit me up and I'll share it.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+    // Each field caps at 3, giving a maximum byte of 15. Observed as 8 -> 9 -> 13 -> 14 -> 15
+    // across four combos, gaining 4 whenever the player's skill completed the combo and 1
+    // whenever the familiar's did. Each spender clears only its own bits: Rally with both fields
+    // full took this from 15 to 3 while TPGauge went 0 -> 250.
+    [FieldOffset(0x10)] public byte InstinctState;
+
+    public BeastmasterKinType KinshipKinType => (BeastmasterKinType)(KinshipState >> 4);
+    public byte KinshipBattlehorn => (byte)(KinshipState & 0x0F);
+    public byte NaturalInstinct => (byte)(InstinctState & 0x03);
+    public byte MasteredInstinct => (byte)((InstinctState >> 2) & 0x03);
```
wolfcomp, line 266: All of this could be done with the bitfield attribute and check how it is accessed in the exe as it could reveal more detail about how each field is accessed.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+    // out further combo advancement until it clears. That is why this stays a byte while
+    // CurrentAffinity below is typed as the enum.
+    [FieldOffset(0x0C)] public byte InstinctualComboState;
+    // Affinity of the most recent instinctual skill, replaced by the resulting Sunstrider or
+    // Moonstalker affinity once an intentional combo resolves. Cleared ~7s after the last
+    // skill. Only ever observed holding 0-6.
+    [FieldOffset(0x0D)] public BeastmasterAffinity CurrentAffinity;
+    // Number of instinctual skills chained so far, shown beneath the Inner Compass.
+    // Increments on any instinctual combo, not only intentional (clockwise) ones, and
+    // increases the extra damage dealt.
```
wolfcomp, line 237: Add this as a xml comment instead of something that can't be viewed without having the code up on github or downloaded on your own pc.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+    // pair was clockwise and so formed an intentional combo - the resulting Sunstrider or
+    // Moonstalker affinity ~2.2s later. If the pair was not clockwise it returns to 0 from 7
+    // instead. The 7 is not an affinity: it marks the "Wavering Heart" status (Status row
+    // 4643, "Otherwise engaged. Unable to perform combos with your familiar."), which locks
+    // out further combo advancement until it clears. That is why this stays a byte while
+    // CurrentAffinity below is typed as the enum.
+    [FieldOffset(0x0C)] public byte InstinctualComboState;
+    // Affinity of the most recent instinctual skill, replaced by the resulting Sunstrider or
+    // Moonstalker affinity once an intentional combo resolves. Cleared ~7s after the last
+    // skill. Only ever observed holding 0-6.
```
wolfcomp, line 233: This comment isn't needed

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
+// Beastmaster is a limited job; its gauge drives the Inner Compass UI.
+// Verified at runtime on patch 7.56 by observing JobGaugeManager while playing.
+[StructLayout(LayoutKind.Explicit, Size = 0x18)]
+public struct BeastmasterGauge {
+    // Beastmaster's own TP, 0-250. This does NOT regenerate passively - it is granted by the
+    // combo bonuses on the player's weaponskill chain (Axeblade Bite grants +13, Shieldsplitter
+    // +15), and a broken combo grants nothing. Instinctual skills spend the entire pool.
```
wolfcomp, line 207: This comment isn't needed

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
+// Beastmaster is a limited job; its gauge drives the Inner Compass UI.
+// Verified at runtime on patch 7.56 by observing JobGaugeManager while playing.
```
wolfcomp, line 202: This comment isn't needed

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+[StructLayout(LayoutKind.Explicit, Size = 0x18)]
+public struct BeastmasterGauge {
+    // Beastmaster's own TP, 0-250. This does NOT regenerate passively - it is granted by the
+    // combo bonuses on the player's weaponskill chain (Axeblade Bite grants +13, Shieldsplitter
+    // +15), and a broken combo grants nothing. Instinctual skills spend the entire pool.
+    [FieldOffset(0x08)] public byte TPGauge;
+    // Familiar TP, 0-250. A single pool shared across Battlehorn slots - it is NOT reset by
+    // swapping familiars. Arrives in fixed +12 increments at irregular intervals (observed
+    // 1.3s, 3.1s, 19.2s and 3.0s apart), so it is not a server tick either; it tracks the
+    // familiar's auto-attacks. Familiar actions spend the entire pool.
```
wolfcomp, line 212: This comment isn't needed

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+    [FieldOffset(0x09)] public byte FamiliarTPGauge;
+    // Familiar TP immediately before the most recent familiar action, which scales that
+    // action's potency. A familiar action expends the ENTIRE familiar pool - FamiliarTPGauge
+    // drops to 0 on the same frame - so "TP consumed" and "TP at time of use" are the same
+    // number and this field is both. Seen taking 132 and 168 on actions that zeroed a gauge
+    // holding exactly 132 and 168.
```
wolfcomp, line 218: This comment isn't needed

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+    [FieldOffset(0x0A)] public byte FamiliarTPAtLastUse;
+    // Which Battlehorn slot's familiar is summoned; 0 when none. This is the SLOT, not the
+    // creature - the summoned beast's identity is not exposed here.
+    [FieldOffset(0x0B)] public byte ActiveBattlehorn;
```
wolfcomp, line 222: Suffix the field with Index and remove the comment as it is not needed.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,74 @@ public struct ViperGauge {
...
+    // Reads the active affinity, then 7 while an instinctual combo resolves, then - if the
+    // pair was clockwise and so formed an intentional combo - the resulting Sunstrider or
+    // Moonstalker affinity ~2.2s later. If the pair was not clockwise it returns to 0 from 7
+    // instead. The 7 is not an affinity: it marks the "Wavering Heart" status (Status row
+    // 4643, "Otherwise engaged. Unable to perform combos with your familiar."), which locks
+    // out further combo advancement until it clears. That is why this stays a byte while
+    // CurrentAffinity below is typed as the enum.
+    [FieldOffset(0x0C)] public byte InstinctualComboState;
```
wolfcomp, line 230: Check how this is actually accessed and stored in the exe and remove the comment as it is not needed to be this long.

FFXIVClientStructs/FFXIV/Client/Game/JobGaugeManager.cs
```diff
@@ -39,5 +39,7 @@ public unsafe partial struct JobGaugeManager {
     [FieldOffset(0x08), CExporterUnion("Gauge")] public WarriorGauge Warrior;
     [FieldOffset(0x08), CExporterUnion("Gauge")] public GunbreakerGauge Gunbreaker;
 
+    [FieldOffset(0x08), CExporterUnion("Gauge")] public BeastmasterGauge Beastmaster;
+
```
wolfcomp, line 43: Put this next to the other melee ones instead of as its own.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
```diff
@@ -129,3 +129,40 @@ public enum SerpentCombo : byte {
...
+// intentional combo is simply the next affinity in sequence, wrapping Eldritch back to Volant.
+// Completing one grants a fifth or sixth affinity, which the game calls the "two additional
+// instinctual affinities". Which of the two you get is the parity of the affinity the combo
+// started from: odd -> Sunstrider, even -> Moonstalker. Pairing those two in turn triggers
+// what the game calls an "infinitive combo".
+// The upgrade only lands if the target survives the combo. Kill it with the second skill and
+// you still get Wavering Heart and a ChainCount increment, but no Sunstrider or Moonstalker.
+// These values match the player statuses the job applies, which are contiguous and in this
+// same order: Status rows 4595 Volant Heart, 4596 Rampant Heart, 4597 Durant Heart,
+// 4598 Eldritch Heart, 4599 Sunstrider, 4600 Moonstalker.
```
wolfcomp, line 143: This comment isn't needed

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
```diff
@@ -129,3 +129,40 @@ public enum SerpentCombo : byte {
...
+    Durant = 3,
+    Eldritch = 4,
+    Sunstrider = 5,
+    Moonstalker = 6,
+}
+
+// A familiar's kin type, which decides which Kinship status Borrow grants and therefore which
+// ability Beast Mode turns into. Matches the order of the eight contiguous Kinship statuses,
+// Status rows 4602 Beast Kinship through 4609 Ash Kinship. All eight values observed directly
+// in the gauge, each cross-checked against the Kinship status the familiar actually granted.
```
wolfcomp, line 157: This comment isn't needed

wolfcomp (changes requested):
Still feels like AI written.

redstrate:
On the comment point: comments like that are so hilariously long and novel-ish tend to make the code unreadable instead of understandable. They need to be cut entirely - as suggested already - or written more concisely, CS doesn't need to be a job tutorial. That also applies to the commit message pre and post AI TBH, but I guess that's more of a personal opinion.

The real juice here is how and any quirks of how the data is stored that wouldn't be immediately obvious from the variable name or type alone, what wolfcomp suggests is going to be much better than manual documentation on every little nibble and bit.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
wangel: Fixed --- using the bitfield attribute.  I don't have a decompiler installed, everything was done during runtime in the game.  I will get something installed.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
wangel: I will work on getting a decompiler installed.  Everything I did came from in the game during runtime.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGaugeEnums.cs
```diff
@@ -129,3 +129,27 @@ public enum SerpentCombo : byte {
...
+    None = 0,
+    Beastkin = 1,
+    Vilekin = 2,
+    Cloudkin = 3,
+    Seedkin = 4,
+    Wavekin = 5,
+    Scalekin = 6,
+    Soulkin = 7,
+    Ashkin = 8,
+}
```
wolfcomp: These mirror the exact same values as is in the XBMPet row which is just a byte type. So this can get removed.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,25 @@ public struct ViperGauge {
...
+    [FieldOffset(0x08)] public byte TPGauge;
+    [FieldOffset(0x09)] public byte FamiliarTPGauge;
+    [FieldOffset(0x0A)] public byte FamiliarTPAtLastUse;
+    [FieldOffset(0x0B)] public byte ActiveBattlehornIndex;
+    [FieldOffset(0x0C)] public byte InstinctualComboState;
+    [FieldOffset(0x0D)] public BeastmasterAffinity CurrentAffinity;
+    /// <summary>Instinctual skills chained so far. Shown beneath the Inner Compass. Increases dmg dealt.</summary>
+    [FieldOffset(0x0E)] public byte ChainCount;
+    [BitField<byte>(nameof(KinshipBattlehornIndex), 0, 4)]
+    [BitField<BeastmasterKinType>(nameof(KinshipKinType), 4, 4)]
```
wolfcomp: Change type to byte and rename to `Classification` when removing the enum. To get the names get it from Addon with `17740 + Classification`

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,25 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
+[GenerateInterop]
+[StructLayout(LayoutKind.Explicit, Size = 0x18)]
+public partial struct BeastmasterGauge {
+    [FieldOffset(0x08)] public byte TPGauge;
+    [FieldOffset(0x09)] public byte FamiliarTPGauge;
+    [FieldOffset(0x0A)] public byte FamiliarTPAtLastUse;
+    [FieldOffset(0x0B)] public byte ActiveBattlehornIndex;
```
wolfcomp, line 250: Actual BattlehornIndex is this - 1.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
```diff
@@ -198,6 +198,25 @@ public struct ViperGauge {
     public SerpentCombo SerpentCombo => (SerpentCombo)(SerpentComboState >> 2);
 }
 
+[GenerateInterop]
+[StructLayout(LayoutKind.Explicit, Size = 0x18)]
```
wolfcomp, line 244: Add Inherits with type of JobGauge

wolfcomp (changes requested):
Just some stuff when looking through the exe.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
wangel: Changed and noted.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
wangel: Updated and noted.

FFXIVClientStructs/FFXIV/Client/Game/Gauge/JobGauges.cs
wangel: Done.  Inherited the JobGauge.

PR #1956, opened 2026-09-16
Name IconComponentFlags bit 13 IsImageDisabled

Seeker1437:
Bit 13 is the flag SetIconImageDisableState toggles at 0x140691A40 in 2026.08.11, alongside setting the icon image multiply to 50 or 100.


Haselnussbomber:
Wrong. I'll fix it. Once the servers let me in.

Seeker1437:
Completely wrong? I don't wanna PR anything I misunderstood D:

I felt confident enough I guess when I looked at how it was used

Haselnussbomber:
Yes, a flag was added after IsRecipe so everything shifted.

Seeker1437:
Ah the other pr is likely wrong as well

Seeker1437:
Ah I never went back and rebuilt the db against the new build **headdesk**

PR #1957, opened 2026-09-16
Fix inverted parameter on AtkComponentDragDrop.SetIconDisableState

Seeker1437:
Passing true clears icon flag bit 13 and sets the multiply to 100, passing false sets it and dims to 50 (0x1406D1BE0 in 2026.08.11).


Haselnussbomber:
I renamed it to `SetIconEnabled` in 58b817142b9c1b45462991cfe80a2828df7eaed3 for clarity, but just noticed that name might also causes confusion. /shrug

